PR #8, opened 2015-06-30
Extending ReplyKeyboardMarkup and implementing the offset getUpdates parameter

pevdh:
I made two small commits, one concerning the ReplyKeyboardMarkup class and another to implement the offset parameter in the getUpdates API call (see https://core.telegram.org/bots/api#getupdates)

The ReplyKeyboardMarkup additions made my life a whole lot easier, I just thought I'd share.

The other commit was necessary to fix an internal bug my bot was experiencing, where I would only get up-to-date Updates when I passed the offset parameter along with the getUpdates call. This commit might improve the efficiency of the TeleBot#get_update function as well.

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -42,6 +42,9 @@ class TeleBot:
         self.__stop_polling = False
         self.interval = 3
 
+    def get_last_update_id(self):
+        return self.update_id_list[-1] if len(self.update_id_list) > 0 else None
+
     def get_update(self):
         result = apihelper.get_updates(self.token)
         if result['ok'] is not True:
--- a/telebot/apihelper.py
+++ b/telebot/apihelper.py
@@ -37,10 +37,13 @@ def send_message(token, chat_id, text, disable_web_page_preview=None, reply_to_m
     return check_result(method_url, req)
 
 
-def get_updates(token):
+def get_updates(token, offset=None):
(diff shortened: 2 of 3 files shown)


eternnoir:
Great!! Thanks for your contribution.

eternnoir:
Hello pevhd

I found offset has problem. Use offset the getUpdates API still return all updates. Api doc says "Identifier of the first update to be returned".
I guess it means "get all updates, and set first update update_id with offset number". It will let cache method invalid. So I revert PR#9.

pevdh:
You are completely right, the API doc states: 
"Offset: Identifier of the first update to be returned. Must be greater by one than the highest among the identifiers of previously received updates."
So the offset becomes
get_last_update_id () + 1
Using this method, the message cache becomes unnecessary, as we'll only need to store the id of the last update received. 

My "internal bug" was not actually a bug. The api's 'limit' parameter is set to 100 by default. So should you receive more than 100 messages in less than 24 hours, the last updates won't be added to the getUpdates response. 
There are two ways to solve this problem, either by increasing the 'limit' parameter, or implementing offset in the correct way. I think the last option is most clean, shall I try to fix this tomorrow and submit another pull request?

eternnoir:
OK,thanks your suggestion. I'll wait your PR.

PR #11, opened 2015-06-30
Implemented the offset parameter.

pevdh:
Implemented the offset parameter for the getUpdates API call according to https://core.telegram.org/bots/api#getupdates
In short: the offset is calculated by adding 1 to the id of the last received Update.
Caching messages in a list is now redundant, so the cache is removed. Only the id of the last received Update is kept for future getUpdates calls.

Also did some code clean-up.


eternnoir:
Thanks a lot.

PR #16, opened 2015-07-01
make telebot compatible with python3

ntcong:
small fixes to make telebot compatible with python3. I would prefer using logging, but there's not much use in the module now so maybe later.

also two very small two space PEP8 fixes.


eternnoir:
Thanks for contribution.

PR #21, opened 2015-07-02
Correcting code error

Zerocchi:


pevdh:
Good catch!

Zerocchi:
Because I tried the example and it throws syntax errors haha.

PR #32, opened 2015-07-12
Updated detailed example to use message handlers

LeoDJ:
Now using message handlers instead of the listener for almost everything (except console output)


eternnoir:
Hi @LeoDJ ,

Please send pull request to develop branch. Thx

PR #47, opened 2015-07-19
more efficient

Oleg-MBO:


Kondra007:
From [https://core.telegram.org/bots/api#update](here)

Update
This object represents an incoming update.
update_id	 (integer)	The update‘s unique identifier. Update identifiers start from a certain positive number and increase sequentially. This ID becomes especially handy if you’re using Webhooks, since it allows you to ignore repeated updates or to restore the correct update sequence, should they get out of order.

Issue #51, opened 2015-07-24
Webhook with flask on google app engine, SSL error


a4fr:
I am using pyTelegramBotAPI and flask for handel my telegram bot messages. I'm using the new method `process_new_messages`. Every thing is all right but when i want to send a reply message in `listener` function, `requests` module raise an exception about **SSL connection**:
```
File "/base/data/home/apps/s~mynerdbot/1.385939809966824449/lib/requests/adapters.py", line 429, in send
    raise SSLError(e, request=request)
SSLError: Can't connect to HTTPS URL because the SSL module is not available.
```

I test `urllib` for sent message. It wat ok.
I test `requests` manually and result was **SSLError**.

PR #53, opened 2015-07-24
Append urllib functions to apihelper

a4fr:
Append urllib functions to apihelper for use SSL connection in free Google App Engine.
#issue in #51


pevdh:
Are you sure there isn't any other way to solve your SSLError? Like as suggested here, http://stackoverflow.com/questions/30381581/sslerror-cant-connect-to-https-url-because-the-ssl-module-is-not-available-on.

a4fr:
I can't do this.
I check google app engine manual, SSL request is open just for billing account. So i edit module and now it's working correctly.

PR #55, opened 2015-07-26
Fix video type class

sgomez:
According the API, the Video type has some bugs:

- thumb field is optional
- caption doesn't exists (is a Message field)

Also, there are a typo on 'mime_type'.


eternnoir:
Merged to develop branch.

PR #67, opened 2015-09-03
Check if field is None

Kondra007:
A slight change, still not very good though. Something like `{!s}{!s}{!s}.format(....)` seems a better aproach


pevdh:
Thanks!

PR #79, opened 2015-09-30
fix race-condition for register_for_reply

huiyiqun:
Here is a race-condition in current code.

If `command A` registers `for_reply a`, and `for_reply a` registers `for_reply b`, then `for_reply b` will be wrongly deleted from the list and never be triggered.

Because before `for_reply a` is called, `for_reply a` is located in the first position of `message_subscribers_callbacks`, but after `for_reply a` accomplishes, when `for_reply b` is registered, `for_reply b` is located in the first place and `for_reply a` is located in the second one.
The `for_reply b` is wrongly deleted and `for_reply a` is left.

There are two methods to resolve this problem:
1. Reserve the `message_subscribers_callbacks` and the `message_subscribers_messages`. As a result, the position of `for_reply a` will not change during the process. But considering that when there is too many `for_replys`, the oldest will be removed, the similar problem will happen.
2. Search the index of `for_reply` before deleting it. It's ugly somewhat but it works.

What's more, I'm not sure if there will some other race-conditions cansed by multi-thread, such as `message_subscribers_messages` correspond with an unmeet `message_subscribers_callbacks`.
I add a lock to keep the two list matching. If it's not needed, I will remove it.

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -56,6 +56,7 @@ class TeleBot:
 
         self.message_subscribers_messages = []
         self.message_subscribers_callbacks = []
+        self.message_subscribers_lock = threading.Lock()
 
         # key: chat_id, value: handler list
         self.message_subscribers_next_step = {}
@@ -336,11 +337,12 @@ class TeleBot:
         :param callback:    The callback function to be called when a reply arrives. Must accept one `message`
                             parameter, which will contain the replied message.
         """
-        self.message_subscribers_messages.insert(0, message.message_id)
-        self.message_subscribers_callbacks.insert(0, callback)
-        if len(self.message_subscribers_messages) > 10000:
-            self.message_subscribers_messages.pop()
-            self.message_subscribers_callbacks.pop()
+        with self.message_subscribers_lock:
+            self.message_subscribers_messages.insert(0, message.message_id)
+            self.message_subscribers_callbacks.insert(0, callback)
+            if len(self.message_subscribers_messages) > 10000:
+                self.message_subscribers_messages.pop()
+                self.message_subscribers_callbacks.pop()
 
     def _notify_message_subscribers(self, new_messages):
         for message in new_messages:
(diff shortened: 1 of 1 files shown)


pevdh:
I don't think I understand the race condition here, could you elaborate (with some code perhaps)?

huiyiqun:
I'm sorry for my English.

The problematic code is here:

https://github.com/huiyiqun/zaobot/blob/master/plugins/event.py

It's complex, I will post a simpler code later.

huiyiqun:
Ok, here is the simple code, I do not test it.
```python3
from telebot import TeleBot, types

bot = TeleBot(token)

@bot.message_handler(commands=['hello'])
def hello(message):
    msg = bot.reply_to(message, 'step 1', reply_markup=types.ForceReply())
    bot.register_for_reply(msg, step_one)

def step_one(message):
    msg = bot.reply_to(message, 'step 2', reply_markup=types.ForceReply())
    bot.register_for_reply(msg, step_two)

def step_two(message):
    pass
```

When I say '/hello' to the bot, it asks me to reply to 'step 1'. I answer it and the bot asks me to reply to 'step 2', but `step_two` is never called.

I have read code of `pyTelegramBotAPI`. I find that before `step_one` is called,  [index](https://github.com/eternnoir/pyTelegramBotAPI/blob/036d000b955658feb44638f3fcd8e0eb7cffe4b8/telebot/__init__.py#L352) is 0. While `step_one` is called, `step_two` is registered. As a result, at [here](https://github.com/eternnoir/pyTelegramBotAPI/blob/036d000b955658feb44638f3fcd8e0eb7cffe4b8/telebot/__init__.py#L356), index of `step_one` is 1 and index of `step_two` is 0. So `step_two` is removed but `step_one` is left. So `step_two` will never be triggered.

huiyiqun:
Maybe this is not race-condition, but it's surely a bug.

pevdh:
I see! This definitely is a bug. The fix is to copy https://github.com/eternnoir/pyTelegramBotAPI/blob/master/telebot/__init__.py#L352 to https://github.com/eternnoir/pyTelegramBotAPI/blob/master/telebot/__init__.py#L354.
That is what you did in your fix as well. The Lock however will prevent even more serious bugs such as 
```python
msg_id_list = []
callback_list = []
Thread 1: msg_id_list.insert(0, some_id)
Thread 2: msg_id_list.insert(0, some_other_id)
Thread 2: callback_list.insert(0, some_other_callback)
Thread 1: callback_list.insert(0, some_callback)

>>> print msg_id_list
[some_other_id, some_id]
>>> print callback_list
[some_callback, some_other_callback]
```
Therefore I think this is a nice PR to fix the problem you mentioned as well as the condition described above.

PR #88, opened 2015-10-15
Little code cleanup

mabitte:
* dict.get(key) returns None if key is not in the dictionary
* used the class method's class reference to instantiate type classes

### DIFF
--- a/telebot/types.py
+++ b/telebot/types.py
@@ -88,7 +88,7 @@ class Update(JsonDeserializable):
         obj = cls.check_json(json_type)
         update_id = obj['update_id']
         message = Message.de_json(obj['message'])
-        return Update(update_id, message)
+        return cls(update_id, message)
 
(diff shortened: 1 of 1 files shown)


telebot/types.py
```diff
@@ -377,16 +342,10 @@ def de_json(cls, json_string):
...
-        mime_type = None
-        file_size = None
-        if 'thumb' in obj:
-            thumb = PhotoSize.de_json(obj['thumb'])
-        if 'mime_type' in obj:
-            mime_type = obj['mime_type']
-        if 'file_size' in obj:
-            file_size = obj['file_size']
-        return Video(file_id, width, height, duration, thumb, mime_type, file_size)
+        thumb = obj.get('thumb')
```
eternnoir: thumb must use PhotoSize.de_json. Bug fixed in  e4f799c7.

PR #105, opened 2016-01-11
Added support for statistics tracking using botan.io

aidarbiktimirov:


Kondra007:
+1 using botan with all my bots. It will be very nice to see botan integrated into this lib

pevdh:
Imho this would unnecessarily complicate the TeleBot class.
A better solution might be to separate the botan listener and make it an example bot in e.g. `examples/botan_tracked_bot.py`.

PR #122, opened 2016-03-20
retry network requests until success

### LINKED ISSUE #108: Polling stops if connection breaks mid-request
I experienced the issue that my bot stopped responding to messages after the host machine gets a new IP due to a forced reconnect by my ISP.

I isolated [the responsible code](https://github.com/eternnoir/pyTelegramBotAPI/blob/master/telebot/apihelper.py#L26), I tested it by executing requests manually. 

Specifically the issue is caused by the requests library. If the bot connects to the API with a requested timeout of 20 seconds, and mid these 20 seconds the connection is lost due to the router reconnecting to the ISP and getting a new IP assigned, the request library silently retries to connect. Somehow, this does not work after the IP changed, and therefore the bot is stuck in the call to `requests.request`, which results in the bot not responding to messages, as polling is halted. 

Suggested fix: add the `timeout` parameter to the call to `request` and potentially make the call again if it times out. Suggested value for the [timeout](http://docs.python-requests.org/en/latest/user/advanced/#timeouts) would be something like `(3.5, None)`, as the timeout should only apply to the connect, not the read. For reading, the timeout is handled by the API.

The linked paragraph even mentions that it is recommended to set a timeout:

> Most requests to external servers should have a timeout attached, in case the server is not responding in a timely manner. Without a timeout, your code may hang for minutes or more.

In regards to testing: I tested the `requests.request` in an interactive python shell to pinpoint the issue, but I could not test the suggested edits in context with the rest of the library. This is due to my inexperience with the package system in python; specifically I don't know how to clone the library, modify it and use it with my current application, and I couldn't find the time to digg into this further. That's why I decided to post my findings so far.

Cheers

ratijas:
fixes #108

### DIFF
--- a/telebot/apihelper.py
+++ b/telebot/apihelper.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 import requests
+import time
 import telebot
 from telebot import types
 from telebot import util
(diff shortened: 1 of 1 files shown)


telebot/apihelper.py
```diff
@@ -126,9 +166,10 @@ def get_updates(token, offset=None, limit=None, timeout=None):
         payload['offset'] = offset
     if limit:
         payload['limit'] = limit
-    if timeout:
+    if timeout is not None:
```
ratijas: by the way, `timeout=0` is perfectly legal value, and telegram API says that in such case it will be 'normal, short polling', i.e. response will be given immediately.

PR #170, opened 2016-05-22
Added edit_date to Message object

Kondra007:
As of Bot API update 2.1, `Message` object now has optional `edit_date` field.

### DIFF
--- a/telebot/types.py
+++ b/telebot/types.py
@@ -179,7 +179,7 @@ class Message(JsonDeserializable):
         if 'reply_to_message' in obj:
             opts['reply_to_message'] = Message.de_json(obj['reply_to_message'])
         if 'edit_date' in obj:
-            opts['edit_date'] = Message.de_json(obj['edit_date'])
+            opts['edit_date'] = obj.get('edit_date')
         if 'text' in obj:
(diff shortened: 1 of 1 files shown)


eternnoir:
edit_date is Integer not Message.

Kondra007:
@eternnoir Sorry, fixed.

pevdh:
Another suggestion, perhaps `obj.get(key)` should be changed to `obj[key]`, because of consistency?

Kondra007:
@pevdh If key doesn't exists, `js["key_which_not_exists"]` will raise `KeyError`, while `js.get("key_which_not_exists")` will simply return `None`. 
Isn't that better?

pevdh:
Yes, you are right once again!

Kondra007:
Maybe we should replace all `obj["key"]` expressions with `obj,get("key")` in future to avoid raising KeyErrors

pevdh:
We should raise KeyErrors if the Telegram API omits  'required' (i.e. always present) keys, imo.

Kondra007:
Good point.

eternnoir:
Merged in feature-changes20160522

PR #171, opened 2016-05-22
Added edited_message to Update object

Kondra007:
As of Bot API update 2.1, Update object now has optional edited_message field.


pevdh:
Fails because the test should be altered, see [test_types.py](https://github.com/eternnoir/pyTelegramBotAPI/blob/master/tests/test_types.py#116). Could you please update the tests?

Kondra007:
@pevdh Nope, that wasn't tests' problem, it was me, who forgot to add `edited_message` to `return cls(...)` expression, so that one argument was missing in `__init__()` function.

Travis build should go fine now.

eternnoir:
Merged in feature-changes20160522

PR #172, opened 2016-05-22
Added "user" to MessageEntity object

Kondra007:
As of Bot API update 2.1, `MessageEntity` object now has optional `user` field.  
It is only present when there's a [text mention](https://telegram.org/blog/edit#new-mentions) in message.

### DIFF
--- a/telebot/types.py
+++ b/telebot/types.py
@@ -304,13 +304,15 @@ class MessageEntity(JsonDeserializable):
         offset = obj['offset']
         length = obj['length']
         url = obj.get('url')
-        return cls(type, offset, length, url)
+        user = obj.get('user')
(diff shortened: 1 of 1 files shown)


eternnoir:
user is User object. Must de_json.

Kondra007:
@eternnoir Sorry again, fixed.

eternnoir:
`user = User.de_json(obj.get('user'))` It's not safe. If MessageEntity's user not exist will cause `ValueError("json_type should be a json dict or string.")`. 

Check https://github.com/eternnoir/pyTelegramBotAPI/commit/8c20f6302294ccf0601e9f0f7ea1477fdd7f4003

PR #173, opened 2016-05-22
Added optional connection (aka sending) timeouts to methods that may …

dmcelectrico:
…upload big chunks of data: send_audio, send_voice, send_document, send_sticker and send_video.

Hi! I just ran into an issue while trying to send audio (4MB+) files through send_document method because the 3.5 seconds connection timeout set by default in apihelper.py weren't enough. 
So I added to the methods  send_audio, send_voice, send_document, send_sticker and send_video an optional argument "timeout" so the user can override this value if he or she considers so.

Cheers!


MasterGroosha:
There're reports that `send_photo` can cause this problem as well. Maybe `timeout` should be added to that method as well?

PR #186, opened 2016-06-02
Fix last_update_id is int. Fix yeild updates.

eternnoir:
last_update_id should be int not update instance.
And fix yeild updates.


eternnoir:
@pevdh Please review it.

PR #223, opened 2016-08-22
Add Exception Catching

bace1996:
Add Exception Catching for requests in function _make_request to avoid
bot crash for sometime telegram server returned HTTP 502 Bad Gateway.


eternnoir:
CI Fail.

PR #284, opened 2017-01-30
Non-ASCII chars for filename.

Kuznitsin:
Telegram doesn't accept rfc2231 styled filename. Using utf-8 directly.


ihoru:
I have the same issue! I would be really nice to fix it ;)

PR #319, opened 2017-04-06
Add more accurate control over threads count

Artom-Kozincev:
In the old version you can change the number of threads only from 1 to 2, now you can change it in any ranges.


skif443:
Very important fix. Please merge.

eternnoir:
Thanks.

PR #338, opened 2017-05-19
Video Note support.

Kylmakalle:
Send and receive round video messages.
Support for `send_video_note` method and `video_note` content type.

[Take a look](https://github.com/eternnoir/pyTelegramBotAPI/pull/338/files#diff-9b5a06a5a5b33ef7c3d1268ac38b778aR338) at `length` parameter, I have made some tests and found out that
only numbers in range `(0, 640)` are working, else you will got an error: `{"ok":false,"error_code":400,"description":"Bad Request: wrong video note length"}`, but there are no any restrictions about video note [length](https://core.telegram.org/bots/api#sendvideonote), by the way, i do not see any changes in `video_note` after `length` parameter use. 

Also changing a duration of `video_note` does not affect it in a right way, just check [@videonote](https://t.me/videonote) and [this gist](https://gist.github.com/Kylmakalle/0de6a10a2b12abc994bfd49ef38fff90)
Seems like only timecodes are affected and `duration` is tied to a `video_note` file. [Screenshot](https://cloud.githubusercontent.com/assets/24507532/26255031/1621fc30-3cc1-11e7-8bd0-bf2ffb3d1a86.png)

[New commit](https://github.com/eternnoir/pyTelegramBotAPI/pull/338/commits/3f5596ddce5a94aaf2cfda11139aab9931c74f01#diff-16fc4c0416dfd7dcc37887265594541dR579), action `upload_video_note` does not work properly for Desktop version, but `record_video_note` is OK on both PC and Mobile versions


eternnoir:
Thanks.

PR #355, opened 2017-06-14
Solve multiprocessing problem

kingjan123:
Solve SSLError: [SSL: DECRYPTION_FAILED_OR_BAD_RECORD_MAC] decryption failed or bad record mac (_ssl.c:1754) error when using multiprocessing lib


JrooTJunior:
I think it's bad solution.

Issue #280, opened 2017-01-25
Exception handling best practice


bryzgaloff:
pyTelegramBotAPI version: 2.3.0
OS: Windows 10 / Ubuntu 16.04.1
Python version: 3.5.1

Is there any way to handle exceptions during task execution? As I can see, pyTelegramBotAPI does it itself [here](https://github.com/eternnoir/pyTelegramBotAPI/blob/master/telebot/util.py#L59):
```python
except:
    logger.debug("Exception occurred")
    self.exc_info = sys.exc_info()
    self.exception_event.set()

    if self.exception_callback:
        self.exception_callback(self, self.exc_info)
    self.continue_event.wait()
```
So I had to change this code to the one you can see below to be able to handle exceptions in the custom way (default telebot logging was not informative at all):
```python
except Exception as e:
    logger.error(type(e).__name__ + " occurred, args=" + str(e.args) + "\n" + traceback.format_exc())
    self.exc_info = sys.exc_info()
    self.exception_event.set()

    if self.exception_callback:
        self.exception_callback(self, self.exc_info)
    self.continue_event.wait()
```
I also have to do some more actions while handling different exceptions so I'm wondering whether there is a better way to handle exceptions. Of course it's not really convenient to change the library code on every platform. Please point me to the right direction. Thank you in advance.

PR #359, opened 2017-06-20
Informative exception message for better exception handling

Yolley:
If some exception occure, bot will just show message 'Exception occurred', which is not informative at all and sometimes makes finding mistake pretty hard thing. Based on this suggestion https://github.com/eternnoir/pyTelegramBotAPI/issues/280, i create pull request with needed module (traceback) imported.

### DIFF
--- a/telebot/util.py
+++ b/telebot/util.py
@@ -4,6 +4,7 @@ import threading
 import re
 import sys
 import six
+import traceback
 from six import string_types
 
 # Python3 queue support.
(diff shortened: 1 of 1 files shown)


telebot/util.py
```diff
@@ -56,8 +57,8 @@ def run(self):
                     self.done_event.set()
                 except Queue.Empty:
                     pass
-                except:
-                    logger.debug("Exception occurred")
+                except Exception as e:
+                    logger.error(type(e).__name__ + " occurred, args=" + str(e.args) + "\n" + traceback.format_exc())
```
the31k: This could be replaced with `logging.exception("Error occured")`, which itself will error-log the exception type, message and its traceback. Just for the sake of code simplicity.

PR #361, opened 2017-06-22
json.dumps(allowed_updates) before sending request

ihoru:
Without this code Telegram Bot API will never get a correct json with set up `allowed_updates` field.

Before this code it's easy to check that passing allowed_updates does not work:
1. send `setWebhook` command with passed `allowed_updates = ['message']` (for example)
2. send `getWebhookInfo` command and you will get `allowed_updates: None` in response.
3. That's why bot continues to get updates non-message type.

### DIFF
--- a/telebot/apihelper.py
+++ b/telebot/apihelper.py
@@ -1,5 +1,10 @@
 # -*- coding: utf-8 -*-
 
+try:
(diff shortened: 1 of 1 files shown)


eternnoir:
Thanks.

PR #429, opened 2017-12-04
 Fix for SendLocation with live period

JekaFST:


JekaFST:
Edit and stop live location methods fixes added

PR #444, opened 2018-01-14
Updating send_media_group method

Forden:
Added automated formatting files for send_media_group. User just provide files's names in `media` list (or way to the each file) and they will be automatically uploaded to telegra.ph/upload and will be sent to Telegram using such Bot API feature as uploading data from URL (https://core.telegram.org/bots/api#sending-files) so size limits are the same: 5 MB max size for photos and 20 MB max for other types of content.
Supported file extensions are: 
GIF, JPEG, JPG, PNG and MP4. 
Also added caption support for send_media_group. Now user can provide some text as `caption` param in send_media_group and it will be set as first file descrption. Limits are standart (0-200 characters).

### DIFF
--- a/telebot/apihelper.py
+++ b/telebot/apihelper.py
@@ -141,7 +141,6 @@ def file_to_send(files, text):
         photoExtensions = ['gif', 'jpeg', 'jpg', 'png'] # photo extensions
         videoExtensions = ['mp4'] # video extensions
         for i in files:
-            print(i)
             url = None
             if len(files_to_send) > 0:
                 text = ''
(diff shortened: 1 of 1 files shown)


JrooTJunior:
I think is bad idea for adding that into this library. You use external platform for uploading images.
And... Do you know about PEP-8?

Brixk:
Yes, I use Telegraph, but as I can see, now library don't support any other methods, so I wrote that this is a temporary solution and it should be in TODO list of main developers. But there are some people who need this function and they don't want to write everything by themselves. It just automate uploading processes.
Yes, I know about PEP, but camelCase guarantees that I won't override some other functions (just in case).

eternnoir:
I do not think its a good idea let library depend on other platform. You can make some example let developer handle large file themselves.

PR #462, opened 2018-02-14
Bot Api 3.6

heyyyoyy:
- Supported text formatting in media captions. Specify the desired parse_mode (Markdown or HTML) when you provide a caption. Example:
```python
bot.send_photo(chat_id, photo, caption='<b>bold text</b>, parse_mode='HTML')
```

- In supergroups, if the bot receives a message that is a reply, it will also receive the message to which that message is replying – even if the original message is inaccessible due to the bot's privacy settings. (In other words, replying to any message in a supergroup with a message that mentions the bot or features a command for it acts as forwarding the original message to the bot).

- Added the new field _connected_website_ to Message. The bot will receive a message with this field in a private chat when a user logs in on the bot's connected website using the [Login Widget](https://core.telegram.org/widgets/login) and allows sending messages from your bot.
To handle a message, you need to set `content_types=['connected_website']`. For example:
```python
@bot.message_handler(content_types=['connected_website'])
def response_from_website(message):
    print(message.connected_website)
```

- Added the new parameter _supports_streaming_ to the sendVideo method and a field with the same name to the inputMediaVideo object.
```python
bot.send_video(chat_id, video, supports_streaming=True)
```

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -486,24 +486,25 @@ class TeleBot:
         return apihelper.delete_message(self.token, chat_id, message_id)
 
     def send_photo(self, chat_id, photo, caption=None, reply_to_message_id=None, reply_markup=None,
-                   disable_notification=None):
+                   parse_mode=None, disable_notification=None):
         """
         Use this method to send photos.
         :param disable_notification:
         :param chat_id:
         :param photo:
         :param caption:
+        :param parse_mode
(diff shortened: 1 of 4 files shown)


vb64:
missed parse_mode argument on edit_message_caption call in __init__.py and apihelper.py

heyyyoyy:
@vb64 thanks, i am added

PR #478, opened 2018-03-17
Add step handler backend mechanism; implement memory and redis backends

bakatrouble:
I've encountered a problem: step handlers do not persist between processes (when using webhooks and running multiple uwsgi/gunicorn/etc workers) or restarts (ie. when bot polling crashes because of exceptions or temporary bot API unavailability), so I've implemented external storage (particularly redis, other backends could be implemented as well) for these handlers. 

I cannot figure out one thing: what's the purpose of `pre_message_subscribers_next_step` dict? Nothing but empty dicts get written there.

Also what do you think about my design decisions, maybe something might be improved?

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -1290,6 +1290,12 @@ class TeleBot:
                     self._exec_task(message_handler['function'], message)
                     break
 
+    def add_step_handler(self, callback):
+        return self.step_handler_backend.add_handler(callback)
+
+    def add_step_handlers(self, *callbacks):
+        self.step_handler_backend.add_handlers(*callbacks)
+
 
 class AsyncTeleBot(TeleBot):
     def __init__(self, *args, **kwargs):
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -9,6 +9,8 @@ import six
 
 import logging
 
+from telebot.step_handler_backends import StepHandlerMemoryBackend
+
 logger = logging.getLogger('TeleBot')
 formatter = logging.Formatter(
     '%(asctime)s (%(filename)s:%(lineno)d %(threadName)s) %(levelname)s - %(name)s: "%(message)s"'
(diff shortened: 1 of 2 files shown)


SkymanOne:
It's good. Because I had got problems with this mechanism while using Heroku

bakatrouble:
Oh, I forgot to add usage example, here it is:
```python
from telebot import AsyncTeleBot, types
from telebot.step_handler_backends import StepHandlerRedisBackend

bot = AsyncTeleBot('TOKEN', step_handler_backend=StepHandlerRedisBackend())


@bot.message_handler(['start'])
def step_1(message: types.Message):
    bot.reply_to(message, 'Step 1')
    bot.register_next_step_handler(message, step_2)


@StepHandlerRedisBackend.add_handler
def step_2(message: types.Message):
    bot.reply_to(message, 'Step 2')


if __name__ == '__main__':
    bot.polling(True)
```

bakatrouble:
Default (StepHandlerMemoryBackend) doesn't require registering all handlers so by default code work without any changes

SkymanOne:
@bakatrouble I will fork your changes. Do you test mechanism with webhook in production?

bakatrouble:
@SkymanOne Not yet, just with polling, restarting manually

SkymanOne:
@bakatrouble I see you are from russia, can you give me your telegram link? 😃

bakatrouble:
Added handler registration method to `TeleBot` class, updated usage example here: https://github.com/eternnoir/pyTelegramBotAPI/pull/478#issuecomment-374749630

bedilbek:
@bakatrouble thank you, really important PR, but I noticed one thing. Maybe it is also good idea to add `redis` as an extra package for this library into `setup.py`, so that if anyone wants to use RedisBackend, he can install with command `pip install pyTelegramBotAPI[redis]`

Badiboy:
Re-implemented here https://github.com/eternnoir/pyTelegramBotAPI/pull/789 and merged.

PR #490, opened 2018-04-12
Fixing and upgrading next step and reply handlers. + minor fixes

CoconutWaffle:
Rename **telebot** package to **pytelegrambotapi** becouse lib named **telebot** exists and it raising many errors

Fixes in register_for_reply and register_next_step mechanisms.

Add methods:
|     _register_for_reply_by_message_id_,
|     _register_next_step_handler_by_chat_id_,
|     _clear_reply_handlers_,
|     _clear_reply_handlers_by_message_id_

p.s. All in one commit because of my mistake...


eternnoir:
I don't think change module name from telebot to pytelegrambotapi is a good idea.

CoconutWaffle:
May be you are rigth...

plasticruler:
> register_next_step_handler is not worked with webhook!

It does. You need to give www-data permission to write to the folder defined for that. And when it does I found it gets buggered up during development so I have to clear the folder out. In my case I am using a folder off /tmp. I would like to see the tracking handled by a more robust solution.

fernaoguerra:
hi @plasticruler, can you share how you did it with the /tmp file? Where did you change the directory where .handler-saves is created?
Thanks

PR #516, opened 2018-05-27
_notify_next_handlers drops messages if empty handler list

Badiboy:
After calling
`clear_step_handler(...)`
the code:
`self.next_step_handlers[chat_id] = []`
left the key in next_step_handlers. When a next message arrives, the old handler executes nothing (no handlers), but still remove message from message queue:
`new_messages.pop(i)`

Updated to pop message only when there are real handlers in the list.

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -1155,10 +1155,11 @@ class TeleBot:
             chat_id = message.chat.id
             if chat_id in self.next_step_handlers.keys():
                 handlers = self.next_step_handlers[chat_id]
-                for handler in handlers:
-                    self._exec_task(handler["callback"], message, *handler["args"], **handler["kwargs"])
+                if (handlers):
+                    for handler in handlers:
+                        self._exec_task(handler["callback"], message, *handler["args"], **handler["kwargs"])
+                    new_messages.pop(i)  # removing message that detects with next_step_handler
                 self.next_step_handlers.pop(chat_id, None)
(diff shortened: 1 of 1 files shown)


Badiboy:
Proofs on existing key and empty handler list:
image

Alternatively, the code:
```
    def clear_step_handler(self, message):
        """
        Clears all callback functions registered by register_next_step_handler().

        :param message:     The message for which we want to handle new message after that in same chat.
        """
        chat_id = message.chat.id
        self.clear_step_handler_by_chat_id(chat_id)
```
can be changed to smth like this:
```
if (chat_id in self.clear_step_handler_by_chat_id):
   del self.clear_step_handler_by_chat_id[chat_id]
```
but the commited version is more safe.

Badiboy:
I saw and fix one more issue.
```
    def _notify_next_handlers(self, new_messages):
        i = 0
        while i < len(new_messages):
            message = new_messages[i]
            chat_id = message.chat.id
            if chat_id in self.next_step_handlers.keys():
                handlers = self.next_step_handlers[chat_id]
                if (handlers):
                    for handler in handlers:
                        self._exec_task(handler["callback"], message, *handler["args"], **handler["kwargs"])
(1)                new_messages.pop(i)  # removing message that detects with next_step_handler
                self.next_step_handlers.pop(chat_id, None)
(2)        i += 1
```
If line (1) pops the message, the line (2) will skip the following message.
https://stackoverflow.com/questions/864603/python-while-loop-condition-evaluation

PR #544, opened 2018-07-30
bugfix-async_function_error

kushan-gunasekera:
async keyword use for name of the function, just replace it from async_func()


Badiboy:
This request is out of date, is it?

Badiboy:
Source repositry is dead.

PR #588, opened 2018-10-20
Update web hook example

seeyijie:
Added web hook implementation of an echo bot for pythonanywhere.


Badiboy:
Source repositry is dead.

PR #590, opened 2018-10-23
fix: 'new_chat_participant' => 'new_chat_member'

MrMebelMan:
Update the field name in `telebot_bot.py` example and fix the content_types field.

### DIFF
--- a/examples/telebot_bot/telebot_bot.py
+++ b/examples/telebot_bot/telebot_bot.py
@@ -44,12 +44,12 @@ def on_user_joins(message):
     if not is_api_group(message.chat.id):
         return
 
-    name = message.new_chat_participant.first_name
-    if hasattr(message.new_chat_participant, 'last_name') and message.new_chat_participant.last_name is not None:
-        name += u" {}".format(message.new_chat_participant.last_name)
+    name = message.new_chat_member.first_name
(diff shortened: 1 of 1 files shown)


Badiboy:
image

Seems still wrong?

MrMebelMan:
@Badiboy, nah, this is correct, only the content-type is in plural :+1:

Badiboy:
Lib has the following code:

        self.new_chat_member = None
        self.new_chat_members = None
        ...
        if 'new_chat_member' in obj:
            opts['new_chat_member'] = User.de_json(obj['new_chat_member'])
            content_type = 'new_chat_member'
        if 'new_chat_members' in obj:
            chat_members = obj['new_chat_members']
            nms = []
            for m in chat_members:
                nms.append(User.de_json(m))
            opts['new_chat_members'] = nms
            content_type = 'new_chat_members'
  
But Telegram Bot API has only

new_chat_members | Array of User | Optional. New members that were added to the group or supergroup and information about them (the bot itself may be one of these members)
-- | -- | --

https://core.telegram.org/bots/api#message

I suppose that **new_chat_member** property and content type are outdated and should be removed from the code of the library as unreachable.

MrMebelMan:
Oh. Thanks!

PR #592, opened 2018-11-01
Update ReadMe

Alex-Blade:
`from telebot import apihelper`
`apihelper.proxy = {'http':'http://10.10.1.10:3128'}`

`{ 'http':'10.10.1.10:3128' }`


Badiboy:
I know, that more than a year had passed. But could you please make a revision of the request, if it is still actual or not?

Issue #542, opened 2018-07-27
Bot API 4.0


eternnoir:
- [ ] Added support for editing the media content of messages: added the method editMessageMedia and new types inputMediaAnimation, inputMediaAudio, and inputMediaDocument.
- [ ] Added the field thumb to the Audio object to contain the thumbnail of the album cover to which the music file belongs.
- [ ] Added support for attaching custom thumbnails to uploaded files. For animations, audios, videos and video notes, which are less than 10 MB in size, thumbnails are generated automatically.
- [ ] tg:// URLs now can be used in inline keyboard url buttons and text_link message entities.
- [ ] Added the method sendAnimation, which can be used instead of sendDocument to send animations, specifying their duration, width and height.
- [ ] Added the field animation to the Message object. For backward compatibility, when this field is set, the document field will be also set.
- [ ] Added two new MessageEntity types: cashtag and phone_number.
- [ ] Added support for Foursquare venues: added the new field foursquare_type to the objects Venue, InlineQueryResultVenue and InputVenueMessageContent, and the parameter foursquare_type to the sendVenue method.
- [ ] You can now create inline mentions of users, who have pressed your bot's callback buttons.
- [ ] You can now use the Retry-After response header to configure the delay after which the Bot API will retry the request after an unsuccessful response from a webhook.
- [ ] If a webhook returns the HTTP error 410 Gone for all requests for more than 23 hours successively, it can be automatically removed.
- [ ] Added vCard support when sharing contacts: added the field vcard to the objects Contact, InlineQueryResultContact, InputContactMessageContent and the method sendContact.

PR #620, opened 2019-02-15
sendAnimation

nev3rfail:
Added the method sendAnimation, which can be used instead of sendDocument to send animations, specifying their duration.
https://github.com/eternnoir/pyTelegramBotAPI/issues/542


xueyuanl:
why still not merge these code, I found the master missed this function.

PR #621, opened 2019-02-18
Alter listener decorators to be able to trigger at any event with `text` or `caption`

nev3rfail:
regexp and command handler now work with any message that have text or caption instead of text messages only

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -1476,8 +1476,8 @@ class TeleBot:
     def _test_filter(filter, filter_value, message):
         test_cases = {
             'content_types': lambda msg: msg.content_type in filter_value,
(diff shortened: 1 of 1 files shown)


Badiboy:
Extraction of some bot control routines (especially commands) from message caption looks as some not habitual behaviour within Telegram...

nev3rfail:
Yes, but in can be useful. For example, to create poll with image or audio file. Anyways, these changes I made for myself and decided to return them to upstream. You're the one to decide are they usefull or unnecessary :) 
image

Badiboy:
@nev3rfail I did not evaluate them as necessary or not. :) I just mentioned that it's not the habitual behaviour and may lead to the confusing results for existing bots. 

May be it's reasonable to make this behaviour optional and off by default.

PR #622, opened 2019-02-18
Choosable blocking

nev3rfail:
Allows to pass event to the next handler. User controlled.

If you know that your listener can or will break the loop (i.e return util.TELEBOT_EAT_ALL) than your listener should have `blocking=True`, for example:
```
@bot.message_handler(func=lambda m: True, blocking=True, final=True)
@bot.channel_post_handler(func=lambda m: True, blocking=True, final=True)
def do_something(msg):
    print("Shall not pass.")
    return util.TELEBOT_EAT_ALL #will break the loop even in threaded mode
```
`final=True` will break loop in any case despite default settings and `blocking` param.
Please remember that blocking listeners order is mandatory.

Supports async mode.

Added `final` key to honor default telebot's behaviour on breaking the loop.

`final` defaults to util.TELEBOT_DEFAULT_BREAK.
Set util.TELEBOT_DEFAULT_BREAK=False to not break the loop by default.

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -459,9 +459,11 @@ class TeleBot:
         logger.info('Stopped polling.')
 
     def _exec_task(self, task, *args, **kwargs):
-        if self.threaded:
+        if self.threaded and not kwargs.get('blocking'):
+            kwargs.pop('blocking', None)
             self.worker_pool.put(task, *args, **kwargs)
         else:
+            kwargs.pop('blocking', None)
             return task(*args, **kwargs)
 
     def stop_polling(self):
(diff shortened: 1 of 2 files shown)


Badiboy:
Hi, @natfarleydev .

My proposal is to make your changes more strict:

1. Make **blocking** and **final** positional (with appropriate code updates):
` def _build_handler_dict(handler, blocking = False, final = util.TELEBOT_DEFAULT_BREAK, **filters):`

2. TELEBOT_EAT_NONE - not used and not needed

3. TELEBOT_DEFAULT_BREAK => DEFAULT_BREAK_HANDLING_LOOP

4. TELEBOT_EAT_ALL  => HANDLER_BREAKS_LOOP

**And the general point**
You updated _notify_command_handlers. But what 3 other handlers: **update**, **reply**, **next**?

nev3rfail:
Will do when I'll have some spare time.

PR #664, opened 2019-06-15
Added LoginUrl to types

vovawed:


OslikAi:
The library is dead, we can disperse and go to other libraries.

PR #684, opened 2019-08-12
Update apihelper.py

xxsokolov:
Hi, to indicate a third-party api-url (reverse proxy) added conditions.
Perhaps not the most elegant solution, but this functionality is very lacking.

apihelper.API_URL = "http://reverseproxy.example.com/bot{0}/{1}"


xxsokolov:
I made a mistake in the code

https://github.com/eternnoir/pyTelegramBotAPI/pull/708

PR #686, opened 2019-08-12
 Travis CI: Drop Python versions that are EOL

cclauss:
https://devguide.python.org/devcycle#end-of-life-branches
https://devguide.python.org/#branchstatus


Badiboy:
Seen too late, sorry: already commited. And let's keep 2.7 for some time :)

PR #718, opened 2019-12-10
fix html_text

drforse:
it wasn't replacing <, > and & until first entity and after last entity, because of return before the replacing when entity.type is None


Badiboy:
Fixed here:
https://github.com/eternnoir/pyTelegramBotAPI/pull/539

PR #764, opened 2020-03-02
changed pickle serializer to dill

HolidayMan:
**pickle** can't serialize nested functions and have many other problems. In **dill** many of them are fixed. For instance, if somebody wants to save a decorated function as a `next_step_handler`, it won't be saved because of pickle problems.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,10 @@
+certifi==2019.11.28
+chardet==3.0.4
+dill==0.3.1.1
+idna==2.7
 py==1.4.29
 pytest==3.0.2
 requests==2.20.0
 six==1.9.0
+urllib3==1.24.3
 wheel==0.24.0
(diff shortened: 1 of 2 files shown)


Badiboy:
1. Oh my god. Seriously?!
image

2. Pickle is a system package that is maintained by Python developers and is included in the The Python Standard Library. Dill is package developed by side team, with non-release version 0.3.x. With last commmit half a year ago and 120 issues on git.
I cannot take responsibility to make such changes in library.

PS. You tried to remove pickle for the second time. Do you have some REAL problems with pickle IN THIS BOT or you simply do not like it so much?

HolidayMan:
It's the first time, when I try to remove **pickle** from library and it's not just like that. I have problems with registering nested (decorated) functions as `next_step_handlers`. And the problem is in pickle. Maybe let's find some other solutions of this problem without removing pickle☺

Badiboy:
Sorry for the long delay.

Will this implementation: https://github.com/eternnoir/pyTelegramBotAPI/pull/786 will be OK for your needs? Full backward compatibility and you can use what you want like dill without affecting requirements etc.

PR #767, opened 2020-03-09
Add @bot.poll_handler to be notified of new poll states

Tkachov:
Pretty straight-forward edits, implemented similarly to other handlers. Updated `types.Poll` so it stores all fields passed by Telegram.

Can be used similar to other handlers:
```
@bot.poll_handler(func=lambda call: True)
def handle_vote(poll):
	print poll
```

Didn't update README and didn't write any tests/examples.

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -167,6 +167,7 @@ class TeleBot:
         self.callback_query_handlers = []
         self.shipping_query_handlers = []
         self.pre_checkout_query_handlers = []
+        self.poll_handlers = []
 
         self.threaded = threaded
         if self.threaded:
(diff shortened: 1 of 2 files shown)


Badiboy:
Hi. Is it tested on some running bots?

Tkachov:
Of course. I've tested it on a test bot though, it just was printing `poll` object so I could see it gets the right data.

Tkachov:
If you want to test it a little bit, you can try adding `@pyTelegramBotAPI_test_bot` to a group and using `/poll` command. It creates a poll and then posts a message every time it `@bot.poll_handler` is notified. I'll keep this bot working for a day or two.

Badiboy:
It's a bad practice, but no time currently for this. :( Let's wait for a feedback from users.

PR #785, opened 2020-04-08
Add Middleware Handler 

### LINKED ISSUE #781: Adding Middleware Handler
Adding middleware feature could be really important addition for this library. That would enable to enter some context before entering any handler. As an example this may be setting a language for current session, or setting up user data for current session and so on.

Also, I could help to create PR for this feature if contributors of this library are willing to add something related to this.

But, I also understand that this issue is a duplicate of #185, #184. So, I am curious why those issues were closed without any notice or acknowledgement from the side of the contributors. Maybe we could also discuss those issues here if you do not mind @Badiboy ? 

Any comments or ideas are appreciated.

Thank you

bedilbek:
Closes #781 

This is my initial implementation for Adding Middleware feature. Thus, any comments or ideas are welcome to discuss here.

I hope this feature will help most of the developers to implement different kind of middleware capabilites starting from session creation to localization.

### DIFF
--- a/examples/middleware/i18n.py
+++ b/examples/middleware/i18n.py
@@ -1,0 +1,53 @@
+#!/usr/bin/python
+
+# This example shows how to implement i18n (internationalization) l10n (localization) to create
+# multi-language bots with middleware handler.
+#
+# Note: For the sake of simplicity of this example no extra library is used. However, it is recommended to use
+# better i18n systems (gettext and etc) for handling multilingual translations.
+# This is not a working, production-ready sample and it is highly recommended not to use it in production.
+#
+# In this example let's imagine we want to introduce localization or internationalization into our project and
+# we need some global function to activate the language once and to use that language in all other message
+# handler functions for not repeatedly activating it.
+# The middleware (i18n and l10n) is explained:
+
+import telebot
+from telebot import apihelper
+
+apihelper.ENABLE_MIDDLEWARE = True
+
+TRANSLATIONS = {
+    'hello': {
+        'en': 'hello',
+        'ru': 'привет',
+        'uz': 'salom'
+    }
+}
+
+_lang = 'en'
+
+
+def activate(lang):
+    global _lang
+    _lang = lang
+
+
+def _(string):
+    return TRANSLATIONS[string][_lang]
+
+
+bot = telebot.TeleBot('TOKEN')
+
+
+@bot.middleware_handler(update_types=['message'])
+def activate_language(bot_instance, message):
+    activate(message.from_user.language_code)
+
+
+@bot.message_handler(commands=['start'])
+def start(message):
+    bot.send_message(message.chat.id, _('hello'))
+
+
+bot.polling()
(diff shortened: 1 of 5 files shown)


bedilbek:
Here I am going to give several use cases where this feature could be useful.
For the sake of simplicity I won't use any extra libraries (redis, database access, gettext and etc) and examples given below are just given to explain the usage in detail and it is not recommended to use them right in production!

bedilbek:
Let's imagine we want to introduce localilization or internationalization into our project and we need some global function to activate the language once and to use that language in all other message handler functions not to reapeatedly activate the language. This is possible with this feature and here comes the solution:

```python
import telebot

TRANSLATIONS = {
    'hello': {
        'en': 'hello',
        'ru': 'привет',
        'uz': 'salom'
    }
}

_lang = 'en'


def activate(lang):
    global _lang
    _lang = lang


def _(string):
    return TRANSLATIONS[string][_lang]


bot = telebot.TeleBot('1036993400:AAEBP69pnNCoxYSMggH5DJSGe9qw9dQR6Sw')


@bot.middleware_handler(update_types=['message'])
def activate_language(bot_instance, message):
    activate(message.from_user.language_code)


@bot.message_handler(commands=['start'])
def start(message):
    print('/start lang:', _lang)
    bot.send_message(message.chat.id, _('hello'))

bot.polling()
```
We have written middleware functions which only works for new messages sent to bot and there inside the function we activate the language once by looking language options of the user and anywhere inside other message handlers, when we want to send localized message to the user, correctly selected translation of the message is sent.

bedilbek:
Or let's imagine we want to create a session for each user who communicates with the bot to store different kind of temporary data while session is active. As an example we want to track the state of the user with the help of this session. So, we need a way to store this session data somewhere globally to enable other message handler functions to be able to use it. If you realize, it is really simple and possible with the implementation of this feature and let's see how it will work:

```python
import telebot

INFO_STATE = 'ON_INFO_MENU'
MAIN_STATE = 'ON_MAIN_MENU'

SESSIONS = {
    -10000: {
        'state': INFO_STATE
    },
    -11111: {
        'state': MAIN_STATE
    }
}


def get_or_create_session(user_id):
    try:
        return SESSIONS[user_id]
    except KeyError:
        SESSIONS[user_id] = {'state': MAIN_STATE}
        return SESSIONS[user_id]


bot = telebot.TeleBot('TOKEN')


@bot.middleware_handler(update_types=['message'])
def set_session(bot_instance, message):
    bot_instance.session = get_or_create_session(message.from_user.id)


@bot.message_handler(commands=['start'])
def start(message):
    bot.session['state'] = MAIN_STATE
    bot.send_message(message.chat.id, bot.session['state'])


@bot.message_handler(commands=['info'])
def start(message):
    bot.session['state'] = INFO_STATE
    bot.send_message(message.chat.id, bot.session['state'])

```

Inside a middleware function we assign a new property to a bot instance which will make our session available during the lifetime of the bot and then this property could be used anywhere else for reading and writing purposes.

abdullaev388:
good job!

PR #794, opened 2020-04-15
add Dice and send_dice

drforse:


Badiboy:
Thanx. We'll merge it in several days after we finish with testing and mergeing of https://github.com/eternnoir/pyTelegramBotAPI/pull/789/

PR #797, opened 2020-04-18
enable is_anonymous, allows_multiple_answers and quiz mode

SiriusKoan:
Add three new options for poll.
I also make `test_poll.py` to test the three options.
Thanks for reviewing my pr. It is my first pr, if there is any problem please tell me :pray:.


Badiboy:
Default value for parameters should be None.
` _is_anonymous=None, allows_multiple_answers=None, poll_type=None, correct_option_id=None_`
It is the general approach here. 

Than in function they should be checked for "is None" (because they are bool) to identify if we need add them to message or not.
```
    if is_anonymous is not None:
        payload['is_anonymous'] = is_anonymous
    if allows_multiple_answersis not None:
        payload['allows_multiple_answers'] = allows_multiple_answers
    ...

```
The reason not to add default value in code is to avoid updates if default values will be changed in Bot Api.

SiriusKoan:
I got it.
Thank you for your reminding me. I will fix it soon.

Badiboy:
1. Init was not changed.
image

Badiboy:
image

1. not + (is not) 
2. poll_type is not checked for None.

Badiboy:
image

Where is the second poll type?

SiriusKoan:
> image
> 
> Where is the second poll type?

Sorry, I think it is default value, so I ignore it.
I will add it immediately.

Badiboy:
@SiriusKoan there should be the same pattern as everywhere:

```
if poll_type is not None:
        payload['type'] = poll_type 
        if poll_type == 'quiz' and correct_option_id is not None:
                payload['correct_option_id'] = correct_option_id
   
```

SiriusKoan:
@Badiboy Thank you for your detailed instructions!

Badiboy:
@SiriusKoan I'm very sorry, but we both were very inattentive. All you added is in the Poll class, which is passed as a third parameter to
_send_poll [__init__.py]_
So no need to modify _send_poll [__init__.py]_ header.

But
_send_poll [apihelper.py]_
should be modified.

However I should take a look, what is the correct way: add parameters there or send Poll here, i do not remember the typical way.

Badiboy:
I propose to cancel this pull request and start new from the scratch.
Poll structures needs total refactoring...

Badiboy:
https://github.com/eternnoir/pyTelegramBotAPI/pull/810

Could you please check if everything is fine there?

SiriusKoan:
@Badiboy I have to say sorry for the fact that I didn't read the document completely and waste your time.
I think the PR is okay. I don't find anything inappropriate.

Badiboy:
No good documentation is the root of many troubles.

There is no problem: anyway Poll implementation was messed and inappropriate. Fine that together we made it better.

PR #800, opened 2020-04-20
Bug fix with middleware

ractyfree:
Fixes the bug which was causing other update_types involved in middleware.

**For instance, when update_types argument was set only for 'messages' this bug was causing other updates notifying this handler.**

So update_types argument hasn't been necessary for the decorator.


bedilbek:
Hi @ractyfree.

Thank you for pointing out on the issue. Actually I created this feature, but as I was not still using it, I did not notice the issue.

I created another pull request so it could be closed.

PR #806, opened 2020-04-24
Update __init__.py

noideaw:
added can_invite_users parameter to restrict_chat_member function


Badiboy:
Here you forget to add the parameter description. :(

PR #808, opened 2020-04-24
add requests session for use torpy

pikss86:


Badiboy:
Please, provide any reasoning for this commit.
Description what it is for and how it can be used.
Etc.

pikss86:
Use case

https://github.com/pikss86/TorTelegramBotExample/blob/master/bot.py

PR #811, opened 2020-04-27
update dice as the api has updated

drforse:
add emoji field for dice

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -647,17 +647,18 @@ class TeleBot:
         """
         return apihelper.delete_message(self.token, chat_id, message_id)
 
-    def send_dice(self, chat_id, disable_notification=None, reply_to_message_id=None, reply_markup=None):
+    def send_dice(self, chat_id, emoji=None, disable_notification=None, reply_to_message_id=None, reply_markup=None):
         """
         Use this method to send dices.
         :param chat_id:
+        :param emoji:
         :param disable_notification:
         :param reply_to_message_id:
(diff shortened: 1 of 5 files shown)


Badiboy:
Deat @drforse .

As far as Dice have a separate class, may be it's reasonal to refactor the Dice-relative code to the way it is done with Polls in this commit: https://github.com/eternnoir/pyTelegramBotAPI/commit/bb4f6a71906382eb3f51f3c06e234835b687905d ? To avoid the growth of send_dice parameters. Nobody knows, how many parameters will be added to Dice.

I do not insist (yet), but the case of Poll shows that it may be reasonable from the very beginnig to avoid refactoring and problems with backward compatibility...

drforse:
Hm, I just made the code in the same style as everything else in the library, it would be a great ideia to refactor all the library, but I just don't see, why make a part of code different from the rest of it.
However... I don't get it... How different the way it is done with Polls in this commit you mentioned?like, yeah, I see that it is used Poll-object in in send_poll, but still... No much matter if you will add
```if poll.new_feature:```
or:
```if new_feature```

etc, like... really, it doesn't change things much, if there will be more new things in polls in botapi, there will needed be more new things in polls in the library and code will growth... Or am I just blind?

Badiboy:
> I see that it is used Poll-object in in send_poll
That was the main point.
Ok left "as is", thank you.

Badiboy:
image

Tests failed. Please, fix.

Issue #775, opened 2020-03-25
No support Bot API 4.5 and Bot API 4.6


heyyyoyy:
### **Bot API 4.6**

- Supported Polls 2.0.

- Added the ability to send non-anonymous, multiple answer, and quiz-style polls: added the parameters _is_anonymous, type, allows_multiple_answers, correct_option_id, is_closed_ options to the method _sendPoll_.

- Added the object _KeyboardButtonPollType_ and the field _request_poll_ to the object _KeyboardButton_.

- Added updates about changes of user answers in non-anonymous polls, represented by the object _PollAnswer_ and the field _poll_answer_ in the _Update_ object.

- Added the fields _total_voter_count, is_anonymous, type, allows_multiple_answers, correct_option_id_ to the _Poll_ object.

- Bots can now send polls to private chats.

- Added more information about the bot in response to the _getMe_ request: added the fields _can_join_groups, can_read_all_group_messages_ and _supports_inline_queries_ to the _User_ object.

- Added the optional field _language_ to the _MessageEntity_ object.

### **Bot API 4.5**

- Added support for two new _MessageEntity_ types, _underline_ and _strikethrough_.

- Added support for nested _MessageEntity_ objects. Entities can now contain other entities. If two entities have common characters then one of them is fully contained inside the other.

- Added support for nested entities and the new tags `<u>/<ins>` (for underlined text) and `<s>/<strike>/<del>` (for strikethrough text) in parse mode HTML.

- Added a new parse mode, _MarkdownV2_, which supports nested entities and two new entities `__ `(for underlined text) and `~ `(for strikethrough text). Parse mode Markdown remains unchanged for backward compatibility.

- Added the field _file_unique_id_ to the objects _Animation, Audio, Document, PassportFile, PhotoSize, Sticker, Video, VideoNote, Voice, File_ and the fields _small_file_unique_id_ and _big_file_unique_id_ to the object _ChatPhoto_. The new fields contain a unique file identifier, which is supposed to be the same over time and for different bots, but can't be used to download or reuse the file.

- Added the field _custom_title_ to the _ChatMember_ object.

- Added the new method _setChatAdministratorCustomTitle_ to manage the custom titles of administrators promoted by the bot.

- Added the field _slow_mode_delay_ to the _Chat_ object.

TeeElPrez:
I'm trying to make a poll NOT anonymous, but i can't.. 
"Got an unexpecter keyword argument 'is_anonymous'.
[...]

PR #834, opened 2020-05-09
UPG: Bot API 4.5 support

tohabyuraev:


tohabyuraev:
The last two commits support Bot API 4.5 issue #775

tohabyuraev:
Last commit supports issue #838

Badiboy:
image

Did you really think that if you place arbitrary parameter in class constructor, it will work fine?

It shows that you NEVER run the bot with your updates. You never run any test. 

I appreciate your willing to participate, but I have no time to make a codereview for you. All your pull requests contains elementary bugs that can be discovered by simply run the code.

Badiboy:
image

You do not understand what's written there and how it works. You make blind changes without understanding. The same problem was when you started refactor "to_dic -> to_dict" and break everything.

tohabyuraev:
Yes, there is a flaw. I forgot to add one field to the constructor of the class. I saw this error and wanted to fix it in the next commit.

But as for to_dic -> to_dict, each method was inside a separate class and did not affect the others. And it was an inaccuracy that needed to be fixed.

I see some flaws in the code and would like to fix them. Unfortunately, I'm not perfect.

Badiboy:
Let's try...

PR #843, opened 2020-05-13
add types.Message.argv types.Message.argc

uWynell:
Almost every command in my bot has parameters, and every time I have to implement it manually.
And recently I was thinking, maybe someone needs it too.
That's why my PR is here! I hope it will not be useless.

_Please do not judge strictly and explain what my mistakes are, if any, because this is my first experience of contributing._

### DIFF
--- a/telebot/types.py
+++ b/telebot/types.py
@@ -273,6 +273,8 @@ class Message(JsonDeserializable):
             opts['author_signature'] = obj.get('author_signature')
         if 'text' in obj:
             opts['text'] = obj['text']
+            opts['argv'] = opts['text'].split(" ")
+            opts['argc'] = len(opts['argv'])
             content_type = 'text'
         if 'entities' in obj:
             opts['entities'] = Message.parse_entities(obj['entities'])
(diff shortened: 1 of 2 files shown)


Badiboy:
So the only profit of the PR is that you make an economy on one function call (split)?
`opts['argv'] = opts['text'].split(" ")`
And the second function call (len)?
`opts['argc'] = len(opts['argv'])`
That's all profit?

With this PR you add TWO parameters to EVERY message. You create array of words for EVERY message (what about messsages of 1K, 2K, 4K symbols) to save from one function call on a command?

uWynell:
Thank You for this reproach.
Please be polite.

drforse:
If u want it, consider making something like this
```
def get_args(self):
    split = self.text.split()
    args = split[1:] if len(split) > 1 else None
    return args
```

or 

```
def get_args(self):
    return self.text.partition(' ')[2] or None
```

or

```
def get_args(self):
    return self.text.partition(' ')[2].split() or None
```

PR #848, opened 2020-05-16
Correct processing of bool parameters. Some timeouts added.

Badiboy:


vikingUnet:
thanks, i hope that timeouts will help us!.. But why +10 seconds? What is different between options "timeout" and "connect-timeout" ?

PR #879, opened 2020-06-15
Added sticker_animation

iPurya:
Added new content_type : sticker_animation


Badiboy:
_content_type_ corresponds to the content type of Message object:
https://core.telegram.org/bots/api#message
There are no such content type as "sticker_animation", only sticker is there. 

So I do not see, why "is_animated" field of "sticker" object is not enough...

Badiboy:
Additionally, it will break backward compatibility of the older code.

iPurya:
You right.
i'm very using this content_type to filter messages that i need to handle. so that i requested this pull. i'm closing this.
Thanks.

PR #889, opened 2020-06-22
Update User type

Otxoto:
can_join_groups
can_read_all_group_messages
supports_inline_queries


Otxoto:
Sorry I'm so noob forking in github

PR #908, opened 2020-07-06
Update types.py

iPurya:
Sending stickers everytime have different file_id, so for detecting special stickers we need a unique parameter.

### DIFF
--- a/telebot/types.py
+++ b/telebot/types.py
@@ -2204,6 +2204,7 @@ class Sticker(JsonDeserializable):
         if (json_string is None): return None
         obj = cls.check_json(json_string)
         file_id = obj['file_id']
+        file_unique_id = obj['file_unique_id']
         width = obj['width']
         height = obj['height']
         is_animated = obj['is_animated']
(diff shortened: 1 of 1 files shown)


Badiboy:
Please, fix tests also.

https://travis-ci.org/github/eternnoir/pyTelegramBotAPI/jobs/705556956

iPurya:
It's been a while telegram gives stickers a unique file id and these tests is for before that.

I think this update will be work well.

Badiboy:
I do not ask why tests fails: I know why they fails.

Your update currently brakes commit build test, because code tests are not updated to new required parameter.

So please update code tests also so we'll have successfull build test.

iPurya:
test codes updated.

Badiboy:
https://github.com/eternnoir/pyTelegramBotAPI/pull/908/files

I do not see any changes here.

Badiboy:
I'm not sure that your patch will be aplied, but let's try...

Badiboy:
https://github.com/eternnoir/pyTelegramBotAPI/blob/master/tests/test_types.py

It is not. So please, make a separate update for tests.

PR #916, opened 2020-07-17
Added nested entities from Bot API 4.5

EskiSlav:


Badiboy:
image

What's the reason for changing this comment?

EskiSlav:
> image
> 
> What's the reason for changing this comment?

changed it because in telegram version 4.5 had been added next entities: `strikethrough` and `underline`, so I added  them to the `_sub` dictionary so it works and there is no need to add entities from telegram 4.5 update ...

Badiboy:
This commend tells that, that NESTED entities should be implemented.

image

EskiSlav:
> This commend tells that, that NESTED entities should be implemented.
> 
> image

Thanks for your observation. I'll fix that

PR #924, opened 2020-07-24
caption as first keyword argument (when it exists)

fgallaire:
For a most consistent argument order


Badiboy:
Impossible to change function signatures: it breaks backward compatibility for existing projects. :(

You may be right: it looks more reasonable, but...

_PS: I do not know if it's a good Python practice to pass named parameters in a raw..._

fgallaire:
Warning for API update in changelog:  possible backward compatibility issues

Badiboy:
For me it's a **very weak** reason to make **lot of problems** to lot of library users.

EskiSlav:
Fully agree with **Badiboy**, because my own projects should be rewritten if these changes applied and many thousands of projects too

PR #930, opened 2020-07-31
Minor keyboard code redisign

mrpes:
- add() method now has a row_width parameter, self.row_width is used if row_width is not specified
- add() and row() methods now support auto row splitting
- add() method now also returns self to allow function chaining
- add() method now throw a ValueError exception if row_width >8 (for inline keyboard) or >12 (for reply keyboard), before this telegram just ignored buttons that didn't fit

Also, type-checking moved to utils to keep code consistent

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -18,7 +18,7 @@ console_output_handler = logging.StreamHandler(sys.stderr)
 console_output_handler.setFormatter(formatter)
 logger.addHandler(console_output_handler)
 
-logger.setLevel(logging.ERROR)
+logger.setLevel(logging.WARNING)
 
 from telebot import apihelper, types, util
 from telebot.handler_backends import MemoryHandlerBackend, FileHandlerBackend
--- a/telebot/types.py
+++ b/telebot/types.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 
+import logging
+
 try:
     import ujson as json
 except ImportError:
@@ -9,6 +11,7 @@ import six
 
 from telebot import util
 
+logger = logging.getLogger('TeleBot')
 
 class JsonSerializable(object):
     """
@@ -807,7 +810,8 @@ class ReplyKeyboardRemove(JsonSerializable):
 class ReplyKeyboardMarkup(JsonSerializable):
     def __init__(self, resize_keyboard=None, one_time_keyboard=None, selective=None, row_width=3):
         if row_width>12:
-            raise ValueError('Telegram does not support reply keyboard row width over 12')
+            logger.warning('Telegram does not support reply keyboard row width over 12')
+            row_width=12
 
         self.resize_keyboard = resize_keyboard
         self.one_time_keyboard = one_time_keyboard
@@ -822,15 +826,16 @@ class ReplyKeyboardMarkup(JsonSerializable):
         when row_width is set to 1.
         When row_width is set to 2, the following is the result of this function: {keyboard: [["A", "B"], ["C"]]}
         See https://core.telegram.org/bots/api#replykeyboardmarkup
-        :raises ValueError: If row_width > 12
         :param args: KeyboardButton to append to the keyboard
         :param row_width: width of row
         :return: self, to allow function chaining.
         """
         row_width = row_width or self.row_width
         
+        
         if row_width>12:
-            raise ValueError('Telegram does not support reply keyboard row width over 12')
+            logger.warning('Telegram does not support reply keyboard row width over 12')
+            row_width=12
         
         for row in util.chunks(args, row_width):
             button_array = []
@@ -907,12 +912,12 @@ class InlineKeyboardMarkup(Dictionaryable, JsonSerializable):
         This object represents an inline keyboard that appears
             right next to the message it belongs to.
         
-        :raises ValueError: If row_width > 8
         :return:
         """
         if row_width>8:
-            raise ValueError('Telegram does not support inline keyboard row width over 8')
-
+            logger.warning('Telegram does not support inline keyboard row width over 8')
+            row_width=8
+        
         self.row_width = row_width
         self.keyboard = []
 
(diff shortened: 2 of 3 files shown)


Badiboy:
image
A subject of backward compatibility issue. Any proof?

Badiboy:
image
And for that. :)

Badiboy:
image
Just to discuss: why not to generate list and return it at once instead of yield? Less iterator calls for a simple task...

mrpes:
Well, I cant show you any proofs for 8 buttons in inline keyboard an all, except that on all my devices, telegram wont show buttons that didn't fit. I kinda think that if something does not work on official Telegram clients, this can be a reason to call this not-supported. 

But I can proove that Telegram does not support row_width>12. Lets try to send InlineKeyboard with 32 buttons:
```python
keyboard = InlineKeyboardMarkup(row_width=128)

keyboard.add(*[InlineKeyboardButton(text=str(i), callback_data=str(i)) for i in range(32)])
    
message = bot.send_message(
    chat_id=message.chat.id, 
    text='Test keyboard',
    reply_markup=keyboard)
    
print(message)
```
So what it prints? I'll copy the "reply_markup" entry:
```json
{"inline_keyboard": [[{"text": "0", "callback_data": "0"}, {"text": "1", "callback_data": "1"}, {"text": "2", "callback_data": "2"}, {"text": "3", "callback_data": "3"}, {"text": "4", "callback_data": "4"}, {"text": "5", "callback_data": "5"}, {"text": "6", "callback_data": "6"}, {"text": "7", "callback_data": "7"}, {"text": "8", "callback_data": "8"}, {"text": "9", "callback_data": "9"}, {"text": "10", "callback_data": "10"}, {"text": "11", "callback_data": "11"}]]}
```
Telegram returns the message we sent, and there are only 12 buttons in it. I think Telegram stores Inline keyboard and Reply keyboard in one place, which is why we can send 12 buttons in a row for Inline keyboard, whereas all official clients can only display 8.

The Telegram API documentation says **NOTHING** about the maximum row width, which is weird. But our little experiment shows that Telegram does not support more than 12 buttons in a row, not only for display, but also for... existing? Telegram servers just don't store more than 12 buttons in a row.


And why yield? Because that even not my code, I just needed a way to split list into chunks. And you're right, it's a simple task, so I think we better try not to optimize code, but keep it simple. Of course, we can use list comprehension (copied from the same SO answer):
```python
return [lst[i:i + n] for i in range(0, len(lst), n)]
```
But it looks worse than yield way

Badiboy:
My main concern about 12 / 8 limit was the next.

If library users **for now** does not have exceptions when they post large button row (it doesn't matter if it works fine or not), so the update **will change** their programs behaviour: new exceptions will appear. So generally it is the compatibility issue.

Thus I propose to change exception to warning (logger.warning) and then force the maximal value to 12 / 8. It will slightly change behaviour (more buttons will appear if someone occasionally use 12+ / 8+) but will not break the program execution. And library users may see in logger outout that they have some issues.

What do you think about that?

mrpes:
This is a bug, after all. Telegram doesn't support row_width>12/8, but the library does. We can keep compatibility for those who used a bug, or do everything as it supposed to be, throw an exception. Forcing the maximum value might seem like a good idea, but it causes another problem - unexpected behavior. Yes, a warning will solve it.. Kinda.
This is not consistent. ValueError exist for a reason. If value is wrong, program must throw exception.

Something throws an exception, something only warns the user, and something just normalizes the value and says nothing. Sure, backward compatibility issues are bad, but unexpected behavior and inconsistent code are much worse.

mrpes:
Sometimes we better break some users programs than make a bunch of useless kludges

Badiboy:
No, sorry. I cannot accept exception in the current update: If Telegram does not return an error - the library should not, even unexpectidely after a version update.

For example, in **my personal case** USERS define how many buttons they will have. I do not know what happen in their bots, so MY personal program will UNEXPECTEDLY start throwing exceptions after library update. I know that it will happen, so I'm able to update code in advance. But other users will not. Even if we will write update note: nobody read them. :)

I hardly insist on warning and we can agree FUTURE update to exception after some time / releases happens, so most of users will be able to see warning and update.

> Sometimes we better break some users programs than make a bunch of useless kludges

We do not MAKE them. They already exist. So we will have a slight way to remove them, but not with a shocking therapy. :)

Badiboy:
@mrpes Houston, we have a prolem.
I installed the update to production and face problems.

> 0. 📔 send_menu error.
> Traceback (most recent call last):
>   File "menu_button.py", line 652, in send_menu
>     final_markup = self.get_keyboard(light, user, menu)
>   File "menu_button.py", line 418, in get_keyboard
>     keyboard.row(*key_line)
>   File "/usr/local/lib/python3.6/dist-packages/telebot/types.py", line 867, in row
>     return self.add(args, 12)
>   File "/usr/local/lib/python3.6/dist-packages/telebot/types.py", line 853, in add
>     button_array.append(button.to_dict())
> AttributeError: 'tuple' object has no attribute 'to_dict'

Badiboy:
The same for another call.

> 0. 📔 Start error.
> Traceback (most recent call last):
>   File "/usr/local/lib/python3.6/dist-packages/telebot/types.py", line 974, in row
>     return self.add(args, 8)
>   File "/usr/local/lib/python3.6/dist-packages/telebot/types.py", line 956, in add
>     button_array = [button.to_dict() for button in row]
>   File "/usr/local/lib/python3.6/dist-packages/telebot/types.py", line 956, in <listcomp>
>     button_array = [button.to_dict() for button in row]
> AttributeError: 'tuple' object has no attribute 'to_dict'

Badiboy:
PS. Just an example why we should not do significant behaviour changes:

image

It's just my screen when I updated to last revision.

Badiboy:
Test case that rise the "AttributeError: 'tuple' object has no attribute 'to_dict'" error.
This code works fine with the original library code.

```
try:
    test_keyboard = types.ReplyKeyboardMarkup(resize_keyboard=True, row_width=10)
    for i in range(20):
        key_line = ()
        for b in range(i):
            key_line = key_line + (str(b),)
        test_keyboard.row(*key_line)
except Exception as err:
    print("{}".format(traceback.format_exc(limit=6)))
```

Do not ask "why do you do in this way", currently it does not matter: it worked before, so update breaks compatibility.

@mrpes we need a compatibility fix. :)

PR #934, opened 2020-08-02
Minor keyboard update followup

Badiboy:


Badiboy:
@mrpes - minor update. Is it OK?

mrpes:
Yeah, it makes the code style better and removing magic values, so it's totally ok

PR #935, opened 2020-08-03
Exceptions classes redesign followup

mrpes:
Changes:
- Now API request can return only one exception per time
- ApiTelegramException now has error_code field, which contains telegram api error code
- API methods now return None if there was error while execution
- Bug fixed

### DIFF
--- a/telebot/apihelper.py
+++ b/telebot/apihelper.py
@@ -127,14 +127,15 @@ def _check_result(method_name, result):
     try:
         result_json = result.json()
     except:
-        raise ApiInvalidJSONException(method_name, result)
+        if result.status_code != 200:
+            raise ApiHTTPException(method_name, result)
+        else:
+            raise ApiInvalidJSONException(method_name, result)
+        
     else:    
         if not result_json['ok']:
             raise ApiTelegramException(method_name, result, result_json)
             
-        elif result.status_code != 200:
-            raise ApiHTTPException(method_name, result)
-    
         return result_json
 
 
--- a/telebot/apihelper.py
+++ b/telebot/apihelper.py
@@ -105,7 +105,10 @@ def _make_request(token, method_name, method='get', params=None, files=None):
             timeout=(connect_timeout, read_timeout), proxies=proxy)
     
     logger.debug("The server returned: '{0}'".format(result.text.encode('utf8')))
-    return _check_result(method_name, result)['result']
+    
+    json_result = _check_result(method_name, result)
+    if json_result:
+        return json_result['result']
 
 
 def _check_result(method_name, result):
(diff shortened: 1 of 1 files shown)


Badiboy:
Several questions, as usual :)

1.
image
If the status is not 200 - we cannot guarantee any json in responce. So I gess we still should check status firstly. Only then we can go to json.


2.
image
Are you sure that json **always** contains error_code? I did not check, so I just wont to confirm that you are sure that this code is safe.


3.
image
"else" - what for? We have no chance to pass the exception if json is bad. :)

mrpes:
1. The reason I check for status after cheking for all other exceptions is that Telegram can return status!=200 and json. For example, if we try to access non-existing chat. Telegram returns status 400 and json
Json
So, status!=200 not guarantees that request not returned any json

If Telegram API will return empty page `result.json()` will throw an exception, library will catch it and throw ApiInvalidJSONException
I think we better check if status code!=200 before we raise ApiInvalidJSONException too, but only if `result.json()` throwed an exception
```python
if result.status_code != 200:
    raise ApiHTTPException(method_name, result)
else:
    raise ApiInvalidJSONException(method_name, result)
```
2. It not always contains `error_code`. It contains `error_code` if `result`'s `ok` field is not true. Before raising exception, we check exactly for it. Cite from [Telegram Bot API - Making requests](https://core.telegram.org/bots/api#making-requests):

> The response contains a JSON object, which always has a Boolean field 'ok' and may have an optional String field 'description' with a human-readable description of the result. If 'ok' equals true, the request was successful and the result of the query can be found in the 'result' field. In case of an unsuccessful request, 'ok' equals false and the error is explained in the 'description'. An Integer 'error_code' field is also returned, but its contents are subject to change in the future. Some errors may also have an optional field 'parameters' of the type ResponseParameters, which can help to automatically handle the error.

3. This is mostly style, if there is `else:` it's obvious to everyone, that code will be executed only if we pass `try-except` block without exceptions

Badiboy:
1. 

> So, status!=200 not guarantees that request not returned any json

Bro, it's not serious. :) 
Status!=200 not guarantees that request **will return json**. So if json is not guaranteed - we cannot check it first. Because if you will get 404, for example, you will raise ApiInvalidJSONException, but the real problem is HTTP.

We should deal with it. For example - two-step checking.
If status == 200 - we proceed to json checking.
If status != 200 then we TRY to get json and error_code there. If we succeed - fine, we have error_code, if no - we have no error_code. It's not a subject to ApiInvalidJSONException: status is the main error source, IMHO.


2.
It's Ok, it's my miss, sorry.

mrpes:
Check out new commit. I did this:
1. Get result and try to get json
2. If exception throwed while getting json (returned json is invalid or empty), we check `status_code`
3. If `status_code!=200`, we `raise ApiHTTPException`, else we `raise ApiInvalidJSONException`

Badiboy:
> Check out new commit. I did this
That's it, thank you! 

PS: please, resolve the conflict and we can proceed. And fix a problem with button chunks
Here: https://github.com/eternnoir/pyTelegramBotAPI/pull/930#issuecomment-668482251
And here: https://github.com/eternnoir/pyTelegramBotAPI/pull/930#issuecomment-668483652

mrpes:
Why `InlineKeyboardMarkup.max_row_keys` value is 12? I replaced magic values in `row()` functions with `self.max_row_keys` parameter and now it works wrong
img

_Also, #937 bug is aleready fixed in this pull request_

mrpes:
Am I supposed to change it back to 8? Or keep it 12?

Badiboy:
@mrpes 

> Why InlineKeyboardMarkup.max_row_keys value is 12? 

CopyPaste, sorry.

Badiboy:
Fixed. https://github.com/eternnoir/pyTelegramBotAPI/pull/938

PR #960, opened 2020-08-25
Add last_update_id parameter for constructor

CrafterKolyan:


Badiboy:
Move this parameter to the end of the list please. To save users who MAY pass unnamed parameters to init.

PR #1008, opened 2020-11-05
Added request_acceptable_exceptions

daveusa31:
In my opinion, this is necessary. So as not to catch such common mistakes. In the future, add more exceptions there

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -2,6 +2,7 @@
 from __future__ import print_function
 
 import logging
+import requests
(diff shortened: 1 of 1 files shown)


Badiboy:
image

Infinite loop on bad connection? Seriously?

Badiboy:
You can use
```
RETRY_ON_ERROR
RETRY_TIMEOUT
MAX_RETRIES
```
in apihelper.py to enable retries on send failure.

PR #1019, opened 2020-11-16
add filename to send_document

tcg-ned:
Add filename parameter to send_document()


Badiboy:
1. You canot add new parameter in the middle of existing ones: it may break backward compatibility.

2. I see no reason to set default value, especially "document"... 

3.
image
What's the crap?!

You cannot commit to LIBRARY in such way.

tcg-ned:
> 1. You canot add new parameter in the middle of existing ones: it may break backward compatibility.
> 2. I see no reason to set default value, especially "document"...
> 
> image
> What's the crap?!
> 
> You cannot commit to LIBRARY in such way.

Oops sorry, I get it wrong. But do you know how to change the filename when sending file? @Badiboy

PR #1025, opened 2020-11-29
Bot API 5.0 pinning-unpinning logic update

alexmechanic:
+ add unpin_all_chat_messages() (former unpin_chat_message())
* update unpin_chat_message() (add message_id arg)


Badiboy:
image

Badiboy:
image

message_id is optional. Should have default value (None) for backward compatibility.

alexmechanic:
Fair point

Badiboy:
image

And what about this? Should I say obviously, that this parameter IS NOT USED in the function?

PR #1043, opened 2020-12-15
Add webhook example for Twisted

vixfwis:
I was integrating my bot with Twisted and figured I could wrap a minimal example and post a PR. Tell me if anything needs fixing. Tested on Arch VPS with latest python (3.9) and libraries from pip


Badiboy:
Please, add some introduction in file. What's that, what for, where and how should be used.

vixfwis:
I've looked through other examples and added a short description with some docs links. That should address the issue, if not please elaborate what exactly you'd like to see there

Badiboy:
Thank you!

PR #1049, opened 2020-12-22
fix restrict_chat_member method

twistfire92:
based on [this discription](https://core.telegram.org/bots/api#restrictchatmember)


Badiboy:
image
Everything is fine with restrict_chat_member.

twistfire92:
> https://core.telegram.org/bots/api#restrictchatmember
> image
> Everything is fine with restrict_chat_member.

No, it's not! Look more closely. 

I add JSON-serialized object

Issue #1031, opened 2020-12-03
Where the method "copyMessage"?


o-murphy:
Can't use method "copyMessage" https://core.telegram.org/bots/api#copymessage

arynyklas:
You can't use it bcs, @eternnoir not want to update lib

trikunai:
Looking forward to that method please!

makisukurisu:
> Looking forward to that method please! @eternnoir please

You can make it by yourself if you really need it. It's pretty similar to forward message, but currently will cause errors so you can add try excepts for date and chat, or just simply do not return new message.

```
### apihelper
def copy_message(token, chat_id, from_chat_id, message_id):
    method_url = r'copyMessage'
    payload = {'chat_id': chat_id, 'from_chat_id': from_chat_id, 'message_id': message_id}
    return _make_request(token, method_url, params=payload)
```

```
### init
    def copy_message(self, chat_id, from_chat_id, message_id):

        return types.Message.de_json(
            apihelper.copy_message(self.token, chat_id, from_chat_id, message_id))
## simply remove return here probably will do the trick of not havving to edit next one
```

```
###types
### ONLY if you want to get back your copied message

ND, NC = None, None
        if (json_string is None): return None
        obj = cls.check_json(json_string)
        message_id = obj['message_id']
        from_user = User.de_json(obj.get('from'))
        try:
            date = obj['date']
        except:
            ND = True
        try:
            chat = Chat.de_json(obj['chat'])
        except:
            NC = True
.....

if ND != True or NC != True:
            return cls(message_id, from_user, date, chat, content_type, opts, json_string)
        else:
            return cls(message_id, from_user, None, None, content_type, opts, json_string)
```

kolayne:
> You can't use it bcs, @eternnoir not want to update lib

@arynyklas, what does it mean? Is he against of this change or there is just no PR with it yet?

Badiboy:
Merged into master branch.

PR #1067, opened 2021-01-10
Added copyMessage method

Aazerra:
#1031

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -7,6 +7,8 @@ import sys
 import threading
 import time
 
+from telebot.types import MessageID
+
 logger = logging.getLogger('TeleBot')
 formatter = logging.Formatter(
     '%(asctime)s (%(filename)s:%(lineno)d %(threadName)s) %(levelname)s - %(name)s: "%(message)s"'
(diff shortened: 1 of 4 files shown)


Aazerra:
i added
but idk i did good or not before accept and merge check the code

Badiboy:
```
        return types.Message.de_json(
            ...
```

image

https://core.telegram.org/bots/api#copymessage

Badiboy:
```
def copy_message(self, chat_id, from_chat_id, message_id, caption=None, parse_mode=None, caption_entities=None,
                     reply_to_message_id=None, allow_sending_without_reply=None, reply_markup=None,
                     disable_notification=None, timeout=None):
```

I suppose it's better to keep original parameters order:
image

Badiboy:
```
    if caption_entities is not None:
        payload['caption_entities'] = caption_entities
```

Entities are classes and cannot be passed like this.

image

Badiboy:
```
  if reply_markup is not None:
        payload['reply_markup'] = reply_markup
```

reply_markup cannot be passed like this. Check other methods who have it.

Badiboy:
And no reason to leave "FIXME".

It's just very brief review. Things are not so easy, you should definitely pay attention when commiting to public library.

Aazerra:
Oh sry
I will pay more attention 
I will fix this thing and push it again

Aazerra:
i added to_json to entites but need to added on other things too ( new update of API added entities to too many types )
the function of _covert_entities i created i thing have a bug  on isinstance because i used first element of array
if its wrong please let me know

PR #1078, opened 2021-01-17
hide token from debug logs

ModischFabrications:
Prevent leaks of the bot token by hiding it from the log, learned it the hard way during https://github.com/eternnoir/pyTelegramBotAPI/issues/1058.

### DIFF
--- a/telebot/apihelper.py
+++ b/telebot/apihelper.py
@@ -73,7 +73,7 @@ def _make_request(token, method_name, method='get', params=None, files=None):
     else:
         request_url = "https://api.telegram.org/bot{0}/{1}".format(token, method_name)
 
-    logger.debug("Request: method={0} url={1} params={2} files={3}".format(method, request_url, params, files))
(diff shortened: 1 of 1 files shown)


Badiboy:
Seems too conservative: you may run several bots in one code, so you need to identify them.

I propose to remove the part of token after ":". So you will see bot ID and identify bot, bot not publish token. It's a bit more complex, but seems more convenient. 

_PS. Do not forget that bot id has not fixed length..._

ModischFabrications:
> several bots

good point, I wasn't aware of that. This will still leak bot ids, which might be unwanted for private bot users that rely on "security by obscurity", but I guess it's the lesser evil. Will be done in a moment...

p1ner0:
.env solves the issue :+1: 

```
import environs

env = environs.Env()
env.read_env('.env')

BOT_TOKEN = env('API_KEY')
```

p1ner0:
storing the API_KEY in a external .env file

PR #1081, opened 2021-01-17
Extend readme with more usage hints

ModischFabrications:
I had a few problems while developing my first bot based on this library, these additions would have helped me and might do so for future generations. It's "paying it forward" so to say.


Badiboy:
image

ModischFabrications:
Alright, should be all major points done. I still had to jump a bit between examples, API reference and the readme to find everything during my development, but this should help to make it more accessible without putting too much work into documentation. Feel free to move or change text segments, I just wanted these changes to be non-invasive.

Should I add a link to my bot here or keep it in a separate PR to keep it clean?

PR #1090, opened 2021-01-30
Custom logging level for infinity_polling

Badiboy:


telebot/__init__.py
```diff
@@ -455,17 +455,28 @@ def __notify_update(self, new_messages):
         for listener in self.update_listener:
             self._exec_task(listener, new_messages)
 
-    def infinity_polling(self, timeout=20, long_polling_timeout=20, *args, **kwargs):
+    def infinity_polling(self, timeout=20, long_polling_timeout=20, logger_level=logging.ERROR, *args, **kwargs):
```
dannkunt, line 458: How should logging level be set?
What if somebody not using logging?
Maybe add type hinting for this

telebot/__init__.py
```diff
@@ -455,17 +455,28 @@ def __notify_update(self, new_messages):
...
+        :param logger_level: Custom logging level for infinity_polling logging. None/NOTSET = no error logging
+        """
         while not self.__stop_polling.is_set():
             try:
                 self.polling(none_stop=True, timeout=timeout, long_polling_timeout=long_polling_timeout, *args, **kwargs)
             except Exception as e:
-                logger.error("Infinity polling exception: %s", str(e))
-                logger.debug("Exception traceback:\n%s", traceback.format_exc())
+                if logger_level and logger_level >= logging.ERROR:
+                    logger.error("Infinity polling exception: %s", str(e))
```
dannkunt, line 471: Why not f'string?

dannkunt (changes requested):

telebot/__init__.py
Badiboy: > How should logging level be set?
infinity_polling(..., logger_level=logging.INFO, ...)

> What if somebody not using logging?
¯\_(ツ)_/¯ Use logging...

> Maybe add type hinting for this
I'll add it to parameter description.

telebot/__init__.py
Badiboy: https://blog.pilosus.org/posts/2020/01/24/python-f-strings-in-logging/
(Not the best link, but I cannot find the best one fast)

PR #1101, opened 2021-03-03
reply_to alternative

aiexz:
Tests were passed successfully


Badiboy:
No reason for two identical functions. If you do not like reply_to - make your own decorator for it...

aiexz:
@Badiboy you don't always need to make a literal reply. Just send the message to the same chat

Badiboy:
reply_to and answer_to are synonims. Nobody will fill the difference. 

Also I do not see the so much trouble in 
send_message(message.chat.id, text)
in contrast to
answer_to(message, text)

To economy of 8 symbols by adding special function for that? Looks too expensive. sorry.

PR #1168, opened 2021-06-01
Bug fixed on set_game_score

anvarjamgirov:
fixed wrong ordered argument error on calling apihelper.set_game_score method in set_game_score


Badiboy:
Thanks

PR #1172, opened 2021-06-03
Update util.py

SwissCorePy:
Improved `split_string` function. Splits text into more meaningful parts.

### DIFF
--- a/telebot/util.py
+++ b/telebot/util.py
@@ -17,6 +17,8 @@ try:
 except:
     pil_imported = False
 
+MAX_MESSAGE_LENGTH = 4096
+
 logger = logging.getLogger('TeleBot')
 
 thread_local = threading.local()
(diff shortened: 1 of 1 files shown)


Badiboy:
I think it's not possible to change existing function begaviour to save the backward compatibility of the legacy code behaviour.

It's possible to add your improvements either as a separate funtion or as an optional function behaviour (that will be enabled either with GLOBAL_CONSTANT or with optional parameter). 

Me personally prefer separate function, but the choice is up to you.

SwissCorePy:
> I think it's not possible to change existing function begaviour to save the backward compatibility of the legacy code behaviour.
> 
> It's possible to add your improvements either as a separate funtion or as an optional function behaviour (that will be enabled either with GLOBAL_CONSTANT or with optional parameter).
> 
> Me personally prefer separate function, but the choice is up to you.

It is backward compatible. the only difference ist that `chars_per_string` is now optional

Badiboy:
> It is backward compatible. the only difference ist that chars_per_string is now optional.

No, it will not produce absolutely the same result due to new splitting points ("/n" etc), thus is will not be fully compatible. I'll appreciate new function.

SwissCorePy:
ok i will make it a new function.

PR #1183, opened 2021-06-09
Fixed a bug in `user_link`

SwissCorePy:
`user_link` returned an empty string if `include_id` was set to `False`


Badiboy:
Thanks! AFAIK there was some claim about userlink, but I do not remeber the exact case.

PR #1192, opened 2021-06-19
base_url added

coder2020official:
I've added a little change in base_url. It can be set in TeleBot class. Base_url helps to make a request to another API, local API.
Check tests in the tests folder.
Also, added an example in the example folder.

### DIFF
--- a/examples/base_url.py
+++ b/examples/base_url.py
@@ -1,0 +1,7 @@
+import telebot
+bot = telebot.TeleBot("token",base_url="http://localhost:8081") #change your base_url. If you want to make request to api.telegram.org, leave it.
+
+@bot.message_handler(commands=['start'])
+def welcome(message):
+    bot.send_message(message.chat.id,"Hello world")
+bot.polling()
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -92,12 +92,13 @@ class TeleBot:
         setMyCommands
         answerInlineQuery
         """
-
     def __init__(
-            self, token, parse_mode=None, threaded=True, skip_pending=False, num_threads=2,
+            self, token, base_url="", parse_mode=None, threaded=True, skip_pending=False, num_threads=2,
             next_step_backend=None, reply_backend=None, exception_handler=None, last_update_id=0,
             suppress_middleware_excepions=False
     ):
+        if base_url:
+            apihelper.API_URL = base_url
         """
         :param token: bot API token
         :param parse_mode: default parse_mode
--- a/telebot/apihelper.py
+++ b/telebot/apihelper.py
@@ -69,7 +69,8 @@ def _make_request(token, method_name, method='get', params=None, files=None):
     :return: The result parsed to a JSON dictionary.
     """
     if API_URL:
-        request_url = API_URL.format(token, method_name)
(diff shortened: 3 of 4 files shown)


Badiboy:
1. What's the reason to do it so complex?
Constructor => Parameter => set API
Why not simply set apihelper.API_URL before create Telebot?

2. This code:
```
         if base_url:
            apihelper.API_URL = base_url
```
is dangerous because may produce side effects. User assign **local paramterer** that will be inobviously assigned to **global variable**.

So, I do not see the any advantages of using
```
bot = telebot.TeleBot("token",base_url="http://localhost:8081")
```
instead of
```
apihelper.APU_URL="http://localhost:8081"
bot = telebot.TeleBot("token")
```

coder2020official:
I've tested to set API_URL of apihelper, it didn't work. So, i added it to telebot class

Badiboy:
Man, how it can "didn't work", if you do absolutely the same thing now?
image

Badiboy:
Ah, I know why it didn't work for you.
`request_url = API_URL.format(token, method_name)

You should provide URL it appropriate format:
apihelper.API_URL = "http://localhost:8081/bot{0}/{1}"

And everything will be fine. If you have an **issue** - you should raise an **issue**, not a pull request. :)

coder2020official:
But it would be better to add base url that can be set in TeleBot class.
But, adding with format options didn't work as i tried before. It returned:
http://localhost:8071/bot0/1
Instead of formatting options. I added the 'f' before string, no change.
I will retry after some time. Thank you

coder2020official:
image
image
Look here)

coder2020official:
I meant this also. So my pull request is useful 😄

Badiboy:
Feel the difference between what I wrote:
`apihelper.APU_URL="http://localhost:8081"`
and what you wrote:
`apihelper.APU_URL=f"http://localhost:8081"`

coder2020official:
image
image
No, still

Badiboy:
You assign the string. Then print it. What you extected to see there?!

Are we playing in kindergarden?

coder2020official:
Oops

coder2020official:
But ptb has base url if i am not mistaken.. it wouldn't be bad if we d add

Badiboy:
Here:
https://github.com/eternnoir/pyTelegramBotAPI/pull/1192#issuecomment-864384814
i mentined 2 points why this update have no value and is dangerous.

SwissCorePy:
> I've tested to set API_URL of apihelper, it didn't work. So, i added it to telebot class

You have to set it to `http://localhost:PORT/bot{0}/{1}`
This works :)

SwissCorePy:
> image
> image
> Look here)

it does not work because you set apihelper.API_URL to "httP://localhost:PORT/bot0/1" beacause you make an expression (f-string). But this string is formated later by making a request usin API_URL.format(token, method_name). 
just delete the f before your string and it should work

coder2020official:
Yes, this is fixed now, but it'd be easier to add it as base_url :/

SwissCorePy:
> Yes, this is fixed now, but it'd be easier to add it as base_url :/

It makes no big difference but would be a good option too :)

PR #1194, opened 2021-06-19
Some small changes

SwissCorePy:
* Fixed type warnings in some editors by changing `var: Type = None` to `var: Union[Type, None] = None`
* changed some args from `obj['arg']` to `obj.get('arg')` if arg is optional
* better PEP-8 compliance for less weak warnings
* added tests for the new type `ChatInviteLink`

I hope the type warnings are gone for you too

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -9,6 +9,7 @@ import time
 import traceback
 from typing import List, Union
 
+# this imports are used to avoid circular import error
 import telebot.util
 import telebot.types
 
@@ -23,7 +24,7 @@ logger.addHandler(console_output_handler)
 
 logger.setLevel(logging.ERROR)
 
-from telebot import apihelper, types, util
+from telebot import apihelper, util, types
 from telebot.handler_backends import MemoryHandlerBackend, FileHandlerBackend
 
 """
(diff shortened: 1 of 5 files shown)


Badiboy:
image
But why?

Badiboy:
image
Location in Venue is not optional:
https://core.telegram.org/bots/api#venue

Badiboy:
image
photos in UserProfilePhotos also is not optional:
https://core.telegram.org/bots/api#userprofilephotos

SwissCorePy:
> image
> But why?

I thougt its not really necessary to make typehints for functions that only got called internally. i thougt i remove them for now.

Badiboy:
> I thougt its not really necessary to make typehints for functions that only got called internally. i thougt i remove them for now.

Ook

SwissCorePy:
> image
> photos in UserProfilePhotos also is not optional:
> https://core.telegram.org/bots/api#userprofilephotos

I know, but i dont know what happens if a user has no profile pictures. i thougt a short check can not hurt :)

SwissCorePy:
> image
> Location in Venue is not optional:
> https://core.telegram.org/bots/api#venue

ok, this is really not not necessairy.. my bad

Badiboy:
image
You seems right, really very strange contruction... "pop" replace correct naming to incorrect one... I think everything should be renamed to option_ids according to https://core.telegram.org/bots/api#polloption, without any "pop"s.

Badiboy:
The remaining updates looks fine, thank you. 

Should I merge this and you will fix updated in new one, or you will update this one?

SwissCorePy:
> image
> You seems right, really very strange contruction... "pop" replace correct naming to incorrect one... I think everything should be renamed to option_ids according to https://core.telegram.org/bots/api#polloption, without any "pop"s.

Yea I thougt i do not touch that for now for beckwards compatibility. so i just left a comment. the pop is to avoid passing unused vars

SwissCorePy:
> The remaining updates looks fine, thank you.
> 
> Should I merge this and you will fix updated in new one, or you will update this one?

I will fix that in a new one. should i change the `option_ids` to the correct name too?

Badiboy:
> Yea I thougt i do not touch that for now for beckwards compatibility. so i just left a comment. the pop is to avoid passing unused vars

I think it's definitely will not work anyway, because it passes to json "options_ids":
 'options_ids': self.options_ids}
which will not be recognized by Telegram. So there is no compatibility which we can keep :)

Badiboy:
> I will fix that in a new one. should i change the `option_ids` to the correct name too?

But "option_ids" is the correct naming:
https://core.telegram.org/bots/api#polloption
I do not understand what you are talking about.

SwissCorePy:
> > I will fix that in a new one. should i change the `option_ids` to the correct name too?
> 
> But "option_ids" is the correct naming:
> https://core.telegram.org/bots/api#polloption
> I do not understand what you are talking about.

Yea you are right. I meant if someone tries to get polloption.options_ids in his code it will not work in the future. but i'll fix that

Badiboy:
> Yea you are right. I meant if someone tries to get polloption.options_ids in his code it will not work in the future. but i'll fix that

Yes, it will not. But this code will not work anyway :)

PR #1195, opened 2021-06-19
Added class ChatMemberUpdated

SwissCorePy:
* Added class `ChatMemberUpdated` to `types.py`
* Simplified `de_json` functions in `WebhookInfo` and `Update` classes (for overall more consistent code)
* changed `options_ids` to `option_ids` in class `PollAnswer`
* Added test for `ChatMemberUpdated` class in `test_types.py` and added the fields `my_chat_member` and `chat_member` to `Update` class and its tests

My last PR for today, i promise ^^

### DIFF
--- a/telebot/types.py
+++ b/telebot/types.py
@@ -94,23 +94,24 @@ class Update(JsonDeserializable):
     def de_json(cls, json_string):
         if json_string is None: return None
         obj = cls.check_json(json_string)
-        update_id = obj['update_id']
-        message = Message.de_json(obj.get('message'))
-        edited_message = Message.de_json(obj.get('edited_message'))
-        channel_post = Message.de_json(obj.get('channel_post'))
-        edited_channel_post = Message.de_json(obj.get('edited_channel_post'))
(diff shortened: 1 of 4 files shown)


Badiboy:
image

Not suitable update without default parameters values ("None") in __init_.

Badiboy:
image

Frankly speaking, this update does not look reasonable. All _init_ parameters are mandatory, you simply put parameters into dict and later unroll dict to parameters... In spite of all other updates, this one looks a bit confusing instead of being more clear and code reduction...

Badiboy:
It's not a key point, just some subjective oppinion for the next PR...

SwissCorePy:
> image
> 
> Frankly speaking, this update does not look reasonable. All _init_ parameters are mandatory, you simply put parameters into dict and later unroll dict to parameters... In spite of all other updates, this one looks a bit confusing instead of being more clear and code reduction...

I know this change is little bit pointless in that class. I did it just for consistency. But why not just modify the dict and pass that to _init_ instead of store all the values in new variables and pass them in the correct order. I like the dict method more. But if you wish i can change it back :)

PR #1198, opened 2021-06-21
Added some missing features

SwissCorePy:
* Added some missing classes, functions and variables from TelegramBotAPI 4.5-5.2 to pyTelegramBotAPI
* Added type hints for (almost) all TeleBot functions

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 from __future__ import print_function
+from datetime import datetime
 
 import logging
 import re
@@ -7,11 +8,13 @@ import sys
 import threading
 import time
 import traceback
-from typing import List, Union
+from typing import Any, Callable, List, Optional, Union
 
 # this imports are used to avoid circular import error
 import telebot.util
 import telebot.types
+
+
 
 logger = logging.getLogger('TeleBot')
 formatter = logging.Formatter(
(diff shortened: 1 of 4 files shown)


Badiboy:
Bro, if you will not stop - we'll never publish the update :)))

Badiboy:
`thumb: Union[Any, str]`

What's the reason of typing "Any"? Especially "Any" or "str"?

SwissCorePy:
> `thumb: Union[Any, str]`
> 
> What's the reason of typing "Any"? Especially "Any" or "str"?

its a placeholder for now. i think we can use bytes or _pyio.BufferedReader.

> Bro, if you will not stop - we'll never publish the update :)))

Haha, sorry should i slow down a little bit? ^^

Badiboy:
image
Ops

Badiboy:
image

I suppose that boolean values should be checkd on None obviously:
`if not (allow_sending_without_reply is None):`
Otherwise we will loose "False" values. It MAY work if the default value on the Telegram side is False, but we cannot be sure what's going on there. So I propose to have obvious check for bools

Badiboy:
Ufff, I spend an hour+ just to review it. Huge work, bro!

SwissCorePy:
> image
> 
> I suppose that boolean values should be checkd on None obviously:
> `if allow_sending_without_reply is not None:`
> Otherwise we will loose "False" values. It MAY work if the default value on the Telegram side is False, but we cannot be sure what's going on there. So I propose to have obvious check for bools.
> 
> image
> Like here. :)

oh oops yes you're right, ill fix that

SwissCorePy:
> image
> Ops

xD typed that name to often to today ^^ is fixed

SwissCorePy:
okey, the bool values in all `to_dict` functions are now checked `if not None`.
But i wait untill you've reviewed the current PR. because also added the class`InputInvoiceMessageContent` and support for tgs stickers.

Badiboy:
I'm done with this one.

SwissCorePy:
ok perfect :) i do new PR after finsishing the tests

PR #1200, opened 2021-06-21
Fix long string blocking version of python on github actions setup

pablodz:
To avoid this

image

that don't show us the version quickly.


Badiboy:
Thanks.

PR #1201, opened 2021-06-23
Added handlers for `my_chat_member` and `chat_member`

SwissCorePy:
I know, no more updates for now, but i think this small addition was needed.
I hope this is ok for you :)

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -541,7 +541,7 @@ class TeleBot:
             Use logger levels from logging as a value. None/NOTSET = no error logging
         :param allowed_updates: A list of the update types you want your bot to receive.
             For example, specify [“message”, “edited_channel_post”, “callback_query”] to only receive updates of these types. 
-            See util.allowed_updates for a complete list of available update types. 
+            See util.update_types for a complete list of available update types. 
             Specify an empty list to receive all update types except chat_member (default). 
             If not specified, the previous setting will be used.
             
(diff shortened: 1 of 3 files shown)


Badiboy:
image

What's the reason? Why this is better then default behaviour (without allowed_updates)?

SwissCorePy:
> image
> 
> What's the reason? Why this is better then default behaviour (without allowed_updates)?

It's necessary because `chat_member` is not requested by default

Bildschirmfoto 2021-06-23 um 18 27 08

Badiboy:
Ops, that's interesting. I suppose Telegram has the reason to set the default behaviour in that way. So may be we should left the default behaviour also but let user to extend it with chat member?

If you agree, I'll take a look what can be a suitable solution.

Badiboy:
PS: I think I see the reason. If you have a group with 20K users your bot will receive tonns of chat_member updates even if nothing else happen in group and even if you do not need them at all. Your system may have a significant workload for nothing.

SwissCorePy:
Ok :) We could let the user set this as argument in bot.polling()

SwissCorePy:
> PS: I think I see the reason. If you have a group with 20K users your bot will receive tonns of chat_member updates even if nothing else happen in group and even if you do not need them at all. Your system may have a significant workload for nothing.

Yes you are right

SwissCorePy:
I have two solutions for that problem.
1. new bool parameter `allow_member_updates` in bot.polling/bot.infinity_polling (default False)
2. new list parameter `allowed_updates` in bot.polling/bot.infinity_polling (default None)

I would prefer the second solution, because it lets the user exactly define which updates he want to recieve. and i think it is more "future-proof".

What do you think? Or do you have a better solution in mind?

SwissCorePy:
sorry for the more commits, just say it if i have to stop

Badiboy:
> sorry for the more commits, just say it if i have to stop

You should ))))))))))

Badiboy:
> What do you think? Or do you have a better solution in mind?

Looks fine, thank you.

Badiboy:
@SwissCorePy Just FYI. I installed all updates on my production system. If everything will work fine - tomorrow I'll publish the release.

SwissCorePy:
Cool thanks

Badiboy:
image

Ups. I'll fix :)

SwissCorePy:
> image
> 
> Ups. I'll fix :)

whoops sorry ^^

PR #1204, opened 2021-06-27
Change document file_name on send.

floydya:


Badiboy:
```
    if not util.is_string(data):
        file_data = data
        if file_name is not None:
            file_data = (file_name, data)
        files = {data_type: file_data}
```

Could you please provide any references that tuple with file_name and data can be passed to the Telegram? And that this code is conformant with Telegram Bot API? I cannot find anything about it in Bot API documentation.

Badiboy:
OK, I googled about multipart via requests. :) 

I suppose that parameter "file_name" may confuse users. I worry that they will try to use it for sending files by name instead of "data". May be it's reasonable to change it to something like visible_file_name? It should hint user that it's not about data, but about representation?....

Additionally I propose to update
```
        if file_name is not None:
            file_data = (file_name, data)
```
to
```
        if file_name:
            file_data = (file_name, data)
```
Empty file names are also not a case.

PR #1203, opened 2021-06-26
2 of 3 Bot API 5.3 changes

MAIKS1900:
- Personalized Commands for different chats
- Custom Placeholders of input field for ReplyKeyboardMarkup and ForceReply.

PR #1206, opened 2021-06-27
2 of 3 Bot API 5.3 changes

MAIKS1900:
second try)

- Personalized Commands for different chats
- Custom Placeholders of input field for ReplyKeyboardMarkup and ForceReply.

#1203

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -1703,21 +1703,52 @@ class TeleBot:
         """
         return apihelper.delete_chat_photo(self.token, chat_id)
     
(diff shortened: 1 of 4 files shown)


Badiboy:
Looks fine.

I propose to make BotCommandScope class abstract to avoid it's manual instantitation by users. Also it seems good to set a description for this class, mentioning that it is abstract and should not be instantitated?

MAIKS1900:
Added comments and made BotCommandScope class abstract

Badiboy:
Amasing, thank you.

PR #1214, opened 2021-06-30
new deprecated decorator

SwissCorePy:
I think this could be useful in the future :) 
Or should we replace the old deprecated decorator?
what du you think?

### DIFF
--- a/telebot/util.py
+++ b/telebot/util.py
@@ -6,7 +6,7 @@ import threading
 import traceback
 import warnings
 import functools
-from typing import Any, List, Dict, Union
+from typing import Any, Callable, List, Dict, Optional, Union
 
 import queue as Queue
 import logging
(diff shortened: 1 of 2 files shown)


Badiboy:
I propose simply replace the old "deprecated" one. I definitly sure nobody used it.

And better to add some tests for it.

SwissCorePy:
ok :)

SwissCorePy:
hope this test is ok and work

Badiboy:
Super, thank you.

> hope this test is ok and work
I hope too. I did not see anyhing in output, hope it means it works. :)

SwissCorePy:
> Super, thank you.
> 
> > hope this test is ok and work
> 
> I hope too. I did not see anyhing in output, hope it means it works. :)

the should not be any output. just a message in the logs. but i dont know where the logs are stored to be honest ^^

SwissCorePy:
But i did a test with printing and it works fine :)

Issue #1224, opened 2021-07-10
'TeleBot' object has no attribute 'worker_pool'


eking-go:
1. What version of pyTelegramBotAPI are you using?

```
pyTelegramBotAPI    3.8.1
```

2. What OS are you using?
```
# cat /etc/debian_version 
10.10
# uname -a
Linux hostname 5.10.0-0.bpo.7-amd64 #1 SMP Debian 5.10.40-1~bpo10+1 (2021-06-04) x86_64 GNU/Linux
```

3. What version of python are you using?
```
Python 3.7.3 (default, Jan 22 2021, 20:04:44) 
[GCC 8.3.0] on linux
```

When I use `threaded=False` in init:
```
bot = telebot.TeleBot(tg_token, threaded=False)
```

I am getting error `'TeleBot' object has no attribute 'worker_pool'` when stops bot:
```
bot.stop_bot()
```

I think it happens there: https://github.com/eternnoir/pyTelegramBotAPI/blob/master/telebot/__init__.py#L743

If I do not use `stop_bot()` or not use `threaded=False` - all ok. All I need from the bot - send a message, without polling (so. I do not need to receive messages - only send). But as I realize threads do not end correctly, because without `threaded=False` for my process raised a number of threads. I see it in systemd status of process:
```
# systemctl status tbot
● tbot.service - TBot service
   Loaded: loaded (/usr/local/bin/tbot.service; enabled; vendor preset: enabled)
   Active: active (running) since Sat 2021-07-10 12:23:37 EEST; 17min ago
 Main PID: 23777 (python3)
    Tasks: 1 (limit: 4460)
   Memory: 17.0M (high: 128.0M max: 164.0M)
   CGroup: /system.slice/tbot.service
           └─23777 python3 /usr/local/bin/tbot.py
```
`Tasks: 1` - this value increasing. In the case of  `threaded=False` value is not increasing

PR #1226, opened 2021-07-11
Get attribute worker_pool of self or value False

David256:
When #1224 happens the attribute `worker_pool` never is defined, then not can be read in https://github.com/eternnoir/pyTelegramBotAPI/blob/db2accc2f84feb594a2f743e8a13cfe6fc16da41/telebot/__init__.py#L743.

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -740,7 +740,7 @@ class TeleBot:
 
     def stop_bot(self):
         self.stop_polling()
-        if self.worker_pool:
+        if getattr(self, 'worker_pool', False):
             self.worker_pool.close()
 
     def set_update_listener(self, listener):


Badiboy:
As far as:
https://github.com/eternnoir/pyTelegramBotAPI/blob/38319871e640721f65b9ac47d64b9e0651d3ddc5/telebot/__init__.py#L209

I propose this solution:
```
    def stop_bot(self):
        self.stop_polling()
        if self.threaded and self.worker_pool:
            self.worker_pool.close()
```

What do you think about it?

David256:
It looks better

Badiboy:
https://github.com/eternnoir/pyTelegramBotAPI/pull/1230

PR #1237, opened 2021-07-19
Replace for loops with comprehensions

monosans:


BLUE-DEVIL1134 (approved):
This seriously wont reduce the execution time but lesser codes is good...

Badiboy:
Thanks.

PR #1254, opened 2021-08-03
🐛 Bugfix: Add return statement to to_dict() method inside InputInvoiceMessageContent type

snikidev:
This is a fix for 
```
A request to the Telegram API was unsuccessful. Error code: 400. Description: Bad Request: can't parse inline query result: Field "input_message_content" must be of type Object
```
error. It happens when you try to send `InputInvoiceMessageContent` as a part of `InlineQueryResultArticle` query with 
```
bot.answer_inline_query()
```


Badiboy:
Thank you.

Badiboy:
Thank you.

PR #1263, opened 2021-08-14
Create chat_member_example.py

coder2020official:
Example of using chatMember


Badiboy:
Thanks.

PS: I cannot check this example, so it's on your responsibility. :)

coder2020official:
It has worked. Update was made by @SwissCorePy, I followed his steps.

PR #1264, opened 2021-08-15
skip_updates

coder2020official:
Added skip_updates
It skip old updates except last one.
Added only to polling and infinity polling. Added example too.
As telegram api says in getUpdates method:
The negative offset can be specified to retrieve updates starting from -offset update from the end of the updates queue. All previous updates will forgotten.
I passed negative 1 as offset
Results are shown in images
Before skip updates
Screen Shot 2021-08-15 at 11 45 19
After 
Screen Shot 2021-08-15 at 11 45 53
As I understood, by passing negative offset will clear all updates except the latest one.

### DIFF
--- a/examples/skip_updates_example.py
+++ b/examples/skip_updates_example.py
@@ -1,0 +1,15 @@
+import telebot
+
+bot = telebot.TeleBot("TOKEN")
+
+@bot.message_handler(commands=['start', 'help'])
+def send_welcome(message):
+	bot.reply_to(message, "Howdy, how are you doing?")
+
+@bot.message_handler(func=lambda message: True)
+def echo_all(message):
+	bot.reply_to(message, message.text)
+
+bot.polling(skip_updates=True) # Will skip old messages when skip_updates is set
+
+# Also, you can use skip_updates in infinity_polling()
--- a/examples/skip_updates_example.py
+++ b/examples/skip_updates_example.py
@@ -1,6 +1,6 @@
 import telebot
 
-bot = telebot.TeleBot("TOKEN",skip_pending=True)# Skip pending skips old updates
+bot = telebot.TeleBot("TOKEN")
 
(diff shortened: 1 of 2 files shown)


Badiboy:
1.
image
That's not exactly what you did (can be used only on bot creation, but is very simillar.

2.
image
There is a method to skip updates. It's better to use it or update it if you'll find that it's not suitable.

coder2020official:
Ok, i will recommit today

coder2020official:
Do you think i need to delete skip_pending and instead put it in polling as skip_update?
I think putting in polling method would be better. Then i will rewrite method you have shown to skip updates

coder2020official:
Removed skip_pending and put instead skip_updates in polling method.

Badiboy:
> Do you think i need to delete skip_pending and instead put it in polling as skip_update?

We cannot delete :)

Badiboy:
> Do you think i need to delete skip_pending and instead put it in polling as skip_update?

We cannot delete existing parameters. :) Compatibility. :(

Additionally I propose to name parameter "skip_pending" to be common to one in Init.

coder2020official:
Aiogram has skip_updates=True in executor, i was getting some things a from there
It is better to put this parameter in polling, isn't it?

coder2020official:
So, what do you think of skip_updates in polling method?

coder2020official:
Made like you said

coder2020official:
I have a second pull request, can u merge or comment this

Badiboy:
> So, what do you think of skip_updates in polling method?

Have nothing against it: why not to have it in both places and let user use what he prefer.

But in my oppinion skip_pending naming in polling/infinity_polling is preferrable to be consistent with skip_pending in Init... What do you think?

coder2020official:
Updated, now it can be used in 2 places

Badiboy:
Amazing. :)

PR #1266, opened 2021-08-16
Update __init__.py

coder2020official:
I added some useful methods. I wanted to separate handlers in other files. But, there was no method to add handler. In order to add handler, I needed to build dict from handler, then use add handler method. I created special methods for it, now, you just create a function with message attribute, and register it.

Screen Shot 2021-08-16 at 22 44 34

What do you think of this change? It would make easier to separate handlers in other files and call them, after register.

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -594,7 +594,7 @@ class TeleBot:
         This allows the bot to retrieve Updates automagically and notify listeners and message handlers accordingly.
 
         Warning: Do not call this function more than once!
-
+        
         Always get updates.
         :param interval: Delay between two update retrivals
         :param none_stop: Do not stop polling when an ApiException occurs.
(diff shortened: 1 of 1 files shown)


coder2020official:
If you accept this PR and testing, you could release a new version. What do you think?

Badiboy:
There is no problem in releasing new version nearly any time :)

coder2020official:
Ok, i will wait :)

Badiboy:
@coder2020official As far as I got - it's the "dynamic" alternative to static handlers adding?
```
@bot.message_handler(commands=['start', 'help'])
def send_welcome(message):
```
If so - why not: more alternatives - more flexible tool :)

coder2020official:
Yes, it will help to register handlers from another files too.

PR #1277, opened 2021-08-24
A Google Cloud Functions webhook

fgallaire:
Working great now

### DIFF
--- a/telebot/util.py
+++ b/telebot/util.py
@@ -440,3 +440,18 @@ def deprecated(warn: bool=False, alternative: Optional[Callable]=None):
         return wrapper
     return decorator
 
+
+# Cloud helpers
+def webhook_functions(bot, request):
+    """A webhook endpoint for Google Cloud Functions FaaS."""
+    if request.is_json:
+        try:
+            request_json = request.get_json()
+            update = types.Update.de_json(request_json)
+            bot.process_new_updates([update])
+            return ''
+        except Exception as e:
+            print(e)
+            return 'Bot FAIL', 400
+    else:
+        return 'Bot ON'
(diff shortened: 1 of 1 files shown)


fgallaire:
Now working great

Badiboy:
You have two pull requests. What is the difference? Or close the outdated if any.

fgallaire:
> You have two pull requests. What is the difference? Or close the outdated if any.

Now only one, this one !

The code is working great now, I use it.

Badiboy:
1. Why the function is named

> webhook_functions

while it can be used **only for Google Cloud Functions**? If this, it should have name like
_webhook_google_cloud_
or something like that.

2. How this function can/should be used? Is it obvious for ones who use Google? (I did not use it, so it's a question)

Badiboy:
image

It's enormous that different function returns have different signatures.

fgallaire:
> image
> 
> It's enormous that different function returns have different signatures.

It's Flask API, the most used web framework in the Python world.

Badiboy:
Okay, thank you.

fgallaire:
>     1. Why the function is named
> 
> 
> > webhook_functions
> 
> while it can be used **only for Google Cloud Functions**? If this, it should have name like
> _webhook_google_cloud_
> or something like that.

The name of the service is _Functions_ and so the name of the function. As it's a generic name, I agree something more is needed, as a service form the _Google Cloud_, I think add just _google_ is shorter and so better.

Done.

>     1. How this function can/should be used? Is it obvious for ones who use Google? (I did not use it, so it's a question)

Yes it's quite obvious, the function do all the boilerplate.

```
def webhook(request):
    util.webhook_google_functions(bot, request)
```

fgallaire:
> Okay, thank you.

Thank you a lot for your work.

PR #1293, opened 2021-09-07
Use native requests/urllib3 retry mechanism

ofen:
https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#urllib3.util.Retry

### DIFF
--- a/telebot/apihelper.py
+++ b/telebot/apihelper.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-import time
 from datetime import datetime
 
 try:
@@ -8,7 +7,8 @@ except ImportError:
     import json
 
 import requests
-from requests.exceptions import HTTPError, ConnectionError, Timeout
+from requests.adapters import HTTPAdapter
+from requests.packages.urllib3.util.retry import Retry
 
 try:
     # noinspection PyUnresolvedReferences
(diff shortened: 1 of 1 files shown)


Badiboy:
Any reasoning of this update? What's wrong with existing implementation?

Badiboy:
1. Debug information is dropped.
2. Session handling is dropped.

It needs valuable argument for such update.

ofen:
Hi, @Badiboy. This one reduce delays after network error/re-connection and works better on slow networks (e.g. cellular).

coder2020official:
Is this still valid?

ofen:
@coder2020official hi! The issue still persists. Current retry implementation may hang on unstable network connection and recovers with 10-40s delays.

coder2020official:
I highly recommend using retry_on_error. This helped me a lot. Even when I put PC into sleep mode, it didn't stop.
`
telebot.apihelper.RETRY_ON_ERROR = True
bot.polling()
`

coder2020official:
Also:
How can I handle reocurring ConnectionResetErrors?
Bot instances that were idle for a long time might be rejected by the server when sending a message due to a timeout of the last used session. Add `apihelper.SESSION_TIME_TO_LIVE = 5 * 60` to your initialisation to force recreation after 5 minutes without any activity.

Badiboy:
@ofen 
https://github.com/eternnoir/pyTelegramBotAPI/pull/1356

Will this way be ok for your proposal and needs?

PR #1295, opened 2021-09-10
is_private filter

coder2020official:
Recently, when I tried to add some filters, you rejected and said, the only is_private can be useful. So I created is_private filter

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,7 +167,7 @@ TeleBot supports the following filters:
 |content_types|list of strings (default `['text']`)|`True` if message.content_type is in the list of strings.|
 |regexp|a regular expression as a string|`True` if `re.search(regexp_arg)` returns `True` and `message.content_type == 'text'` (See [Python Regular Expressions](https://docs.python.org/2/library/re.html))|
 |commands|list of strings|`True` if `message.content_type == 'text'` and `message.text` starts with a command that is in the list of strings.|
-|only_private|Private chat|`True` if chat is private
+|chat_types|list of chat types|`True` if `message.chat.type` in your filter
 |func|a function (lambda or function reference)|`True` if the lambda or function reference returns `True`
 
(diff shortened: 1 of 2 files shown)


Badiboy:
In my oppinion, 
_is_private_
naming is confusing. 

is_private == True - it's clear.

is_private == False - it's confusing.
Does it means NOT PRIVATE (i..e. only groups) or not ONLY private...

May be some other naming? 
"only_private"? for example?

coder2020official:
Ok

Badiboy:
@SwissCorePy - could you please give your oppinion about the naming?

SwissCorePy:
> @SwissCorePy - could you please give your oppinion about the naming?

i recommend making a chat_types filter insted. which defaults to ["private", "group", "supergroup","channel"] when not set.
Eg. 
`@bot.message_handler(chat_types=["private"])`
`@bot.message_handler(chat_types=["group", "supergroup"])`
`@bot.message_handler(chat_types=["channel"])`

coder2020official:
Ok, i will recommit

coder2020official:
Recommited, what do you think?

coder2020official:
I think we don't need default parameter. If it will be None by default, it will work :)
*And, we have another handler for channels, so chat type can't be channel 🤔  @SwissCorePy

SwissCorePy:
Looks good for me :)

> I think we don't need default parameter. If it will be None by default, it will work :)
I mean how it behaves by default if not set.

> *And, we have another handler for channels, so chat type can't be channel 🤔 @SwissCorePy
True. my fault ^^

Badiboy:
> i recommend making a chat_types filter insted

Thanks, bro! Good idea, clear and corresponds well with other handlers.

Badiboy:
@coder2020official thank you! Let's try )

PR #1296, opened 2021-09-11
Filter clearance

Badiboy:
1. Filter optimization: should not store empty filters
2. Filter order: chat_type, content, others
3. Default requests session lifetime set to 600 instead of "forever".
4. Typo fix


Badiboy:
@SwissCorePy could you please review? Especially the first patch of filters.

PR #1298, opened 2021-09-11
Custom filters

coder2020official:
Added new feature - from now you can create your own custom filters.
Here are some examples of how to use it.
image

You can make any check you want. Class should have .check() method, and should have message attribute.
Message update passes to your custom filter, and it runs your filter's check method.

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -3075,10 +3075,9 @@ class TeleBot:
             return filter_value(message)
         else:
             if message_filter in self.custom_filters:
-                if message_filter in self.custom_filters:
-                    filter_check = self.custom_filters.get(message_filter)
-                    if filter_value == filter_check.check(message): return True
-                    else: return False
+                filter_check = self.custom_filters.get(message_filter)
+                if filter_value == filter_check.check(message): return True
+                else: return False
             else:
                 return False
 
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -3074,17 +3074,20 @@ class TeleBot:
         elif message_filter == 'func':
             return filter_value(message)
         else:
-            if message_filter in self.custom_filters:
-                filter_check = self.custom_filters.get(message_filter)
-                if type(filter_value) is bool:
-                    
-                    if filter_value == filter_check.check(message): return True
-                    else: return False
-                else:
-                    if filter_check.check(message,filter_value) is True: return True
-                    else: return False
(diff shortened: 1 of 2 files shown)


coder2020official:
@SwissCorePy @Badiboy What do you think

Badiboy:
image

Badiboy:
```
                if message_filter in self.custom_filters:
                    filter_check = self.custom_filters.get(message_filter)
```
If you already checked, that message_filter is in - you can use []. And less boilerplate code...

```
        elif message_filter == 'func':
            return filter_value(message)
        elif self.custom_filters and message_filter in self.custom_filters:
            filter_check = self.custom_filters[message_filter]
            return filter_value == filter_check.check(message)
        else:
            return False

```

Badiboy:
Function create_filter 
1. No creates but adds
2. not filter but custom filter
Wheat about 
**add_custom_filter**?
(Just for discussion)?

coder2020official:
Hm, i think this sounds better :)

coder2020official:
image

SwissCorePy:
I dont think this is a good idea. because the bot has to do a getChatMember request after every update causing to slow down the bot.

coder2020official:
This is custom filter. Of course it should make request. This is just an example, you can make any filter.

It is exactly the same as any other filter, like let's say content_types.

It checks every time whether content_type is in list of content_types.

I will show another example.

coder2020official:
image

coder2020official:
You can implement any filter for anything - it passes message to your check method. And you perform any check :)

Badiboy:
> I dont think this is a good idea. because the bot has to do a getChatMember request after every update causing to slow down the bot.

This is an example. Responsibility is on filter creator :)

SwissCorePy:
The idea of custom filters is ok. 

but i think the bot should not do requests in the background. (except getting updates)
wouldn't  a custom_message_handler be a better idea?

But im fine with it if you both like it this way :)

Badiboy:
> but i think the bot should not do requests in the background. (except getting updates)

Bro, this example has nothing with proposed update. Just check "Files changed", not the give screens.

SwissCorePy:
Sorry ^^
I think i would never use this feature but i like the idea. :)

coder2020official:
image
Before, it was accepting bool only. Now, i allowed another parameter

coder2020official:
*names of parameters in check method can change. Try to test

PR #1309, opened 2021-09-18
Added new example: anonymous chat-bot

bim-ba:
Hello, i added new example - anonymous chat-bot

With this bot, you will be able to find random users and communicate with them


SwissCorePy:
This is not needed. You can just do `bot.copy_message(...)` instead of all these if/else statements
Bildschirmfoto 2021-09-18 um 19 59 11

PR #1310, opened 2021-09-18
Fixed filter class indexes

Aerodymier:


Badiboy:
Any proofs that original code was incorrect and your update is correct?

Aerodymier:
I was working on a new bot with this API wrapper, I wanted to do custom filters and I went with examples. Pylance wasn't previewing with first one but it previewed with removing the second index in the example. So I removed it and it works very well like that. I haven't tested the one in example but removing it made Pylance preview it + I had no problems.

Aerodymier:
Note: I saw that if custom filter is registered by another script, others can use it with custom_filters.FilterName() so I was wrong and I'm closing this. I still don't know why Pylance doesn't preview like that.

Badiboy:
> I haven't tested the one in example

Commiting to the public library without testing is a very bad practice...

Aerodymier:
> > I haven't tested the one in example
> 
> Commiting to the public library without testing is a very bad practice...

Yeah.. I'm not experienced with public repos, please just ignore it. My bad.

PR #1315, opened 2021-09-25
States, New filter, and more

coder2020official:
- Added 2 examples on: states and admin filter
- Created new filter is_admin
- States as alternative to register_next_step_handler


coder2020official:
Will re-update. Have some issues

coder2020official:
- Implemented custom filters and built-in filters for state handlers

PR #1326, opened 2021-10-06
Added my bots to README

TrevorWinstral:
Only realized this list existed now, otherwise I would have added the Corona bot much earlier.


Badiboy:
You should move data to the end of the list.

PR #1329, opened 2021-10-09
Antiflood

barbax7:
Hi
I created this decorator to avoid TooManyRequest error

User is free to decide if to use this or not with anti_flood parameter in TeleBot.__init__()

In this way, bot can send no more than 100 messages every 30 seconds.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ var/
 
 .idea/
 venv/
+.venv/
 
 # PyInstaller
 #  Usually these files are written by a python script from a template
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -8,6 +8,7 @@ import threading
 import time
 import traceback
 from typing import Any, Callable, List, Optional, Union
+from functools import wraps
 
 # this imports are used to avoid circular import error
 import telebot.util
@@ -144,7 +145,7 @@ class TeleBot:
     def __init__(
             self, token, parse_mode=None, threaded=True, skip_pending=False, num_threads=2,
             next_step_backend=None, reply_backend=None, exception_handler=None, last_update_id=0,
-            suppress_middleware_excepions=False
+            suppress_middleware_excepions=False, anti_flood = False
     ):
         """
         :param token: bot API token
@@ -157,6 +158,8 @@ class TeleBot:
         self.update_listener = []
         self.skip_pending = skip_pending
         self.suppress_middleware_excepions = suppress_middleware_excepions
+        self.anti_flood = anti_flood
+        self.last_time = 0
 
         self.__stop_polling = threading.Event()
         self.last_update_id = last_update_id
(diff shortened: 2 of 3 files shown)


coder2020official:
image
What is purpose of this? Why can't you just return f in the end of function?

coder2020official:
image
And, why if anti flood is false, you are still returning f? Why?

coder2020official:
+ Some methods are missing anti flood.
reply_to, edit_message_caption, and etc.
This methods also can get TooManyRequests error :)

Badiboy:
Now from my side.

1. What is the reason for "0.3 second" delay? Why not 0.2? 0.4? 1 or any other number? Any reasoning?
2. _"In this way, bot can send no more than 100 messages every 30 seconds."_ what is the reason of this limit? Why not 50 messages? 150 messages? Any reasoning?
3. Any reason of this limit for different chats? For different chat types? I definitely need the reason for the limit you selected for the implementation.

**Additional.** Your anti-flood will make delay for everything that is sent via bot. If you want to reach the _"In this way, bot can send no more than 100 messages every 30 seconds."_ limit - this method is too straightforward. It will make unnececary delay always when message rate is low and only in rare case of high rate it will save the bot.

barbax7:
> Now from my side.
> 
> 1. What is the reason for "0.3 second" delay? Why not 0.2? 0.4? 1 or any other number? Any reasoning?
> 2. _"In this way, bot can send no more than 100 messages every 30 seconds."_ what is the reason of this limit? Why not 50 messages? 150 messages? Any reasoning?
> 3. Any reason of this limit for different chats? For different chat types?
> 
> I definitely need the reason for the limit you selected for the implementation.
> 
> **Additional.** Your anti-flood will make delay for everything that is sent via bot. If you want to reach the _"In this way, bot can send no more than 100 messages every 30 seconds."_ limit - this method is too straightforward. It will make unnececary delay always when message rate is low and only in rare case of high rate it will save the bot.

1 and 2. Due to [Official Telegram FAQ](https://core.telegram.org/bots/faq#broadcasting-to-users) a bot can only do no more than 30 messages per second. Doing some tests where i sent only 5 messages per second, i got TooManyRequest error when i reached exaclty 100 messages, and Api Response said me to wait 10 seconds. So, a limit is that a bot can send only 100 messages in 30 seconds.
I don't know if this limit is only for me or is for everyone
With a delay of 0.3 it is sure that bot can't send more than 100 messages in 30 seconds.
3. As expleined in the same link of above, a bot can send only 20 messages per minute in the same channel or group. I haven't written a code for that, my code is more general

Replying to your addiction: you are right, I've thought about this too, but for me it was a bit more difficult. My method work but it isn't elastic
If you want I'll try to make it elastic, checking how may calls it is executing.
My only idea is saving number of calls in a variable and using threading.Timer with a time of 30 that reduce by one the count and start every time bot makes a call

Badiboy:
It may be left as some custom solution for less general cases, but you used wrapper, so for every thousands of people who do not need that solution you created additional function call to wrapper. I.e.:
`my_function -> send_message`
now will be
`my_function -> wrapper -> send_message. `

We got additional function call for nothing. :( It looks not too good.

barbax7:
Talking about the limit of 100 messages every 30 seconds, I haven't the possibility to test it with other accounts. Just to remove every doubt, may you please do a while cycle reaching 100 messages in less then 30 seconds, please?

> I need details of your tests, results and other proofs.

I'll try to write a test file with all cases I went through

> It is a good try to make a potentially useful thing, but it looks yet too simple and less general to be practicaly useful for the cases that arize in various circumstance.

I'm not a so able programmer (that is only a hobby), if you prefere we can close the PR and stop, or wait when I'm free from studying

>  but you used wrapper, so for every thousands of people who do not need that solution you created additional function call to wrapper.

When I studied the decorators in python i saw that is important to use a "function inside the function". You too do it with bot.message_handler, decorators with a function inside that calls `decorator` function, that calls the function defined by the user.

coder2020official:
I think we don't need to limit the messages.
Yes, i know limits. But,this error should be handled by the user itself, not library.

PR #1338, opened 2021-10-16
Added description of the ApiTelegramException as attribute of the class

barbax7:
The class ApiTelegramException now has the attribute "description", returned in the response_json. In this way, the error's description is more accessible.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ var/
 
 .idea/
 venv/
+.venv/
 
 # PyInstaller
 #  Usually these files are written by a python script from a template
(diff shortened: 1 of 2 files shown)


coder2020official:
What is the purpose of this? Why can't you catch error by try/except:
```
try:
    some code

except Exception as e:

    #catch
```
image
And this.

barbax7:
It's better to use:
`from telebot.apihelper import ApiTelegramException

try:
    some code
except ApiTelegramException as e:
    if e.description == 'Forbidden: bot was blocked by the user.':
        do something`

It's always better to use specific exceptions
pyTelegramBotApi provides that exceptions (when an error occures it raises ApiTelegramException or other) so qhy not use them in a better way?

the .venv in .gitignore is just for those people like me who create a virtual environment named ".vemv"

coder2020official:
```
try:
    some code

except Exception as e:
    if e == 'Forbidden: bot was blocked by the user.'

```
They are exactly same thing. No need to add it I think

Badiboy:
Thank you.

PR #1341, opened 2021-10-21
fix logger bug

Dimkashow:


Badiboy:
And when you use webhooks - you should handle it when calling process_updates. 
Or also use exception_handler, but try/except around process_updates is more direct way.

For me personally I cover with try/except EVERY handler, it allows me to effectively localise and fix problems, but it's up to you: you can use one handler and long trace to see what was the reason of the error.

PR #1349, opened 2021-11-03
Antiflood

barbax7:
Created ne function "antiflood" that every user could decide to use by importing it from util


Badiboy:
That's fine, thank you.

PR #1356, opened 2021-11-07
RETRY_ENGINE

Badiboy:
Added RETRY_ENGINE var to api_helper.

Added RETRY_ENGINE = 2 based on native "requests" retry mechanism.

### DIFF
--- a/telebot/apihelper.py
+++ b/telebot/apihelper.py
@@ -9,6 +9,7 @@ except ImportError:
 
 import requests
 from requests.exceptions import HTTPError, ConnectionError, Timeout
+from requests.adapters import HTTPAdapter
 
 try:
     # noinspection PyUnresolvedReferences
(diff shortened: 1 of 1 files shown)


jacopotediosi:
Please note that using RETRY_ENGINE 2 ignores the RETRY_TIMEOUT, resending requests immediately.

Badiboy:
Please note that using RETRY_ENGINE 2 retry works as it is defined in requests library.

```
 retry_strategy = requests.packages.urllib3.util.retry.Retry(
            total=MAX_RETRIES,
        )

```

jacopotediosi:
What I mean is that when requests fail, using RETRY_ENGINE 1, there is a delay of RETRY_TIMEOUT seconds between requests, because of the `time.sleep(RETRY_TIMEOUT)` instructions.

Using RETRY ENGINE 2, however, when a request fails it is retried without delay, i.e. ignoring the RETRY_TIMEOUT value.
In my opinion, the implementation of RETRY_ENGINE 2 should be fixed by setting `backoff_jitter=RETRY_TIMEOUT` in the Retry strategy.

Badiboy:
Make a PR proposal, we'll check.

PR #1369, opened 2021-11-23
CallbackData class added

abdullaev388:


coder2020official:
Token 👀

coder2020official:
Are you sure creating a separate file for this is ok? I think it would be better to implement it in util file.

Badiboy:
> Are you sure creating a separate file for this is ok? I think it would be better to implement it in util file.

Why not. The same as states. :)
PS. I'll check later what's that.

coder2020official:
> > Are you sure creating a separate file for this is ok? I think it would be better to implement it in util file.
> 
> Why not. The same as states. :) PS. I'll check later what's that.

You mean custom filters :)

coder2020official (changes requested):
It would be good to have more description on classes and their methods. Just add parameters overview.
I have tested it, great job!

coder2020official (approved):

PR #1376, opened 2021-12-04
Update echo_bot.py

pythobbotluca:


coder2020official:
Nothing wrong with example.

PR #1383, opened 2021-12-07
Bot API 5.5

coder2020official:
December 7, 2021
Bot API 5.5

Bots are now allowed to contact users who sent a join request to a chat where the bot is an administrator with the can_invite_users administrator right – even if the user never interacted with the bot before.
Added support for mentioning users by their ID in inline keyboards. This will only work in Telegram versions released after December 7, 2021. Older clients will display unsupported message.
Added the methods banChatSenderChat and unbanChatSenderChat for banning and unbanning channel chats in supergroups and channels.
Added the field has_private_forwards to the class Chat for private chats, which can be used to check the possibility of mentioning the user by their ID.
Added the field has_protected_content to the classes Chat and Message.
Added the field is_automatic_forward to the class Message.
Note: After this update it will become impossible to forward messages from some chats. Use the fields has_protected_content in the classes Message and Chat to check this.

Note: After this update users are able to send messages on behalf of channels they own. Bots are expected to use the field sender_chat in the class Message to correctly support such messages.

Note: As previously announced, user identifiers can now have up to 52 significant bits and require a 64-bit integer or double-precision float type to be stored safely.

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -1668,8 +1668,7 @@ class TeleBot:
         return apihelper.set_chat_administrator_custom_title(self.token, chat_id, user_id, custom_title)
 
     
-    def ban_chat_sender_chat(self, chat_id: Union[int, str], sender_chat_id: Union[int, str],
-        until_date:Optional[Union[int, datetime]]=None) -> bool:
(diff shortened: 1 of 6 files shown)


coder2020official:
Need to write some tests

PR #1430, opened 2022-01-31
Installing storage and asyncio_storage

barbax7:
With the previews setup function, the directories (packages) storage and asyncio_storage were not installed.

adding packages = find_packages() it will wind all packages (with init.py file)
It also exclude tests and examples from installing
examples should be not installed by default, since it isn't a package, but i added it anyway to be sure.


Badiboy:
Thank you.

PR #1446, opened 2022-02-10
New CallbackData example

abdullaev388:
Created a directory for all CallbackData examples.
News advanced calendar implementation using CallbackData was added.


Badiboy:
Thank you. 

Hope you tested that it's working? :)

PR #1449, opened 2022-02-12
new advanced TextFilter was added && An example demostrating TextFilt…

abdullaev388:
…er usage


coder2020official (changes requested):
Remove your token.
Instead of prints send a message
In filter, instead of checking ignore_case all time, check once.

abdullaev388:
> Remove your token. Instead of prints send a message In filter, instead of checking ignore_case all time, check once.

removed token, modified ignore_case checks but can't send message in poll_handler that handler do not provide chat_id

coder2020official:
I hope you tested?

coder2020official:
Example on asyncio filter is totally wrong

coder2020official:
You can delete asyncio filter example. I will remake it myself.
And for other examples I hope you have tested it.

coder2020official (approved):

PR #1454, opened 2022-02-15
Additional check of the regexp and commands types

Troshchk:
In this pull request, additional checks of the message handler decorator inputs are added.

The regexp type is expected to be string. This is checked and a warning message is displayed if necessary.

The commands input types are checked. The commands input is converted to a list of strings and a warning message is displayed if necessary.

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -2783,7 +2783,7 @@ class TeleBot:
         if content_types is None:
             content_types = ["text"]
 
-        if isinstance(commands, list) or isinstance(commands, set) or isinstance(commands, tuple):
+        if isinstance(commands, (list, set, tuple)):
             if any(not isinstance(item, (str, int, float)) for item in commands):
                 logger.warning("message_handler: commands' filter should be list of strings (commands), unknown type supplied to the 'commands' filter list. Not able to use.")
             elif all(isinstance(item, str) for item in commands):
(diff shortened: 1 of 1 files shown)


Troshchk:
In general, this small non-invasive improvement will make API usage more user-friendly.

As an example, one might want to have a command, which is "4",  and it can be passed to the commands input as 4, not "4". Adding this check will convert it to the required format. Otherwise one will need to do the whole process from modifying to deploying another time, which is unnecessary(this fix handles it) and might be time-consuming.

Also, commands may be supplied as a list and have integer among strings. In this case, the bot will not work and there will be no warning message about it at all.

There is also a check for regexp, what about it?

Badiboy:
When the bot will not work - he will start reading manuals and see what he need to do to make bot working.
https://github.com/eternnoir/pyTelegramBotAPI#message-handlers

This answer is common for commands, numbers, regexps.

The only real common mistake is str instead of [str]. All other checked cases are kind of exotics and/or absolute absence of understanding about what user do.

Badiboy:
Let's ask for additional oppinion by @coder2020official .

coder2020official:
Hello. Warnings are good, but instead, I would recommend to check filter only once, and if it is inappropriate - just display a warning.

Badiboy:
> I would recommend to check filter only once, and if it is inappropriate - just display a warning.

He do it once. On handler adding...

Badiboy:
Okay, if @coder2020official supports the checking - then please adopt the next.

1. No error corrections. Only warnings. Error hiding is a bad practice: user should coorect mistakes himself.
2. According to that, please make more strict checkings (not allowing "if any(not isinstance(item, (str, int, float))" for example).
3. Wrap all checking in a separate private method and call this method everywhere. No need to put all this boilerplate code multiple times.

coder2020official:
If-elses...

Troshchk:
Ok, thanks for the useful information.

I have adopted as @Badiboy mentioned above.

Badiboy:
Thank you. 

Two remarks.

1. May be make this shorter?
image
```
        if commands is not None:
            self.__check_commands_input(commands, "message_handler")
```
What benefit from the larger form?

2. I highly propose to return the "str" autofix for commands. Checks and error-report will stay "as is", but then do the fix. Like
```
        if commands is not None:
            self.__check_commands_input(commands, "message_handler")
            if isinstance(commands, str):
                commands = [commands]
```

Badiboy:
No time to control the pull for so long. I'll fix it.

PR #1511, opened 2022-04-23
Added sync i18n class based middleware

abdullaev388:


Badiboy:
custom_filters.py are old and will be reverted.

PR #1524, opened 2022-04-30
Markdown & Html functions added(Beta version, still in progress)

coder2020official:
## Description
I created a new file which contains all functions that are necessary for markdown and html.
With those functions, you can achieve formatting easier and faster.

## Describe your tests
How did you test your change?
Every function was tested by sending message.
Python version: 3.9.6

OS: MacOs BigSur

## Checklist:
- [x] My changes won't break backend compatibility
- [x] I made changes for async and sync
- [x] Added examples

### DIFF
--- a/docs/source/formatting.rst
+++ b/docs/source/formatting.rst
@@ -1,0 +1,10 @@
+==================
+Formatting options
+==================
+
+
+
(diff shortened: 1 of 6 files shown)


coder2020official:
Also corrected traceback in polling.

PR #1542, opened 2022-05-11
added examples of multibot

abdullaev388:
Examples of multibot (sync and async)


coder2020official (changes requested):
Well done!
Just one thing: is it good to create a new instance for each new webhook request?
image

abdullaev388:
> 

it is not good, but this example is supposed to be as simple as possible.
it can be upgraded. For example instead of saving True as a value of token, the bot instance can be saved there like:
tokens = {}
tokens['token'] = TeleBot()
and then retreived whenever update comes. But anything like that users should implement by themselves

coder2020official:
Let's wait for badiboy's response

Badiboy:
We discussed and desided that for your case (multibots) it's fine. Thanks. :)

PR #1548, opened 2022-05-14
CopyMessage return type fix

AHOHNMYC:
## Description
[CopyMessage](https://core.telegram.org/bots/api#copymessage) returns [MessageId](https://core.telegram.org/bots/api#messageid) while in code it returned `int`

Tests [uses MessageId type already](https://github.com/eternnoir/pyTelegramBotAPI/blob/master/tests/test_telebot.py#L325)

## Checklist:
- [x] My changes won't break backend compatibility
- [x] I made changes for async and sync


Badiboy:
Thanks!

PR #1551, opened 2022-05-15
fix: warning none_stop parameter is deprecated

everpcpc:
## Description
fix warning
```
WARNING: polling: none_stop parameter is deprecated. Use non_stop instead.
```
when using `infinity_polling `

## Describe your tests
How did you test your change?

Python version: 3.9

OS: macos

## Checklist:


Badiboy:
OMG, thank you.

🙈🙈🙈 It was stupid...

PR #1565, opened 2022-05-21
Update create_invite_link.py

DevAdvik:
Added .types while importing inline markup keyboards (fix)
Removed threading import since message is not to be deleted

### DIFF
--- a/examples/create_invite_link.py
+++ b/examples/create_invite_link.py
@@ -20,7 +20,7 @@ def newmember(msg):
     InviteLink = invite.invite_link #Get the actual invite link from 'invite' class
     mrkplink = InlineKeyboardMarkup() #Created Inline Keyboard Markup
     mrkplink.add(InlineKeyboardButton("Join our group 🚀", url=InviteLink)) #Added Invite Link to Inline Keyboard
-    bot.send_message(msg.chat.id, f"Hey there {msg.from_user.first_name}, Click the link below to join our Official Group." reply_markup=mrkplink)
+    bot.send_message(msg.chat.id, f"Hey there {msg.from_user.first_name}, Click the link below to join our Official Group.", reply_markup=mrkplink)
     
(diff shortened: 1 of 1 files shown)


coder2020official:
Remove while true

coder2020official:
You are basically putting while true in while true.

DevAdvik (commented):
Removed while loop

coder2020official:
Import both reply markups on the same line

DevAdvik (commented):
Imported _both_ inline keyboards in same line

DevAdvik (commented):
FINAL FIX

coder2020official:
Another problem:
What about putting reply keyboard markup creation after imports? Creating reply markup for every user is a bad idea.

coder2020official:
And i guess that's it.

DevAdvik:
> What about putting reply keyboard markup creation after imports?
What does that mean? I don't get it:/
Like not importing markup keyboards at all?

coder2020official:
image
This, before handlers, but after imports

DevAdvik:
So like making `mrkplink` inline keyboard right after importing the module? That's neat, thanks!!

DevAdvik (commented):
Assigned mrkplink right after importing markup keyboards

Badiboy:
@coder2020official Put "approve" when it's ready to merge.

coder2020official:
image
Move this line under markup creation as well.

DevAdvik:
This cant be done I guess, the invitelink is created inside the newmember function, so url needs to be assigned there as well

coder2020official:
Sorry for being innatentive.. I did not see that you were creating markup specially for user. So, InlineKeyboardMarkup creation should be in handler to avoid more buttons..

DevAdvik (commented):
Done, added the InlineKeyboardMarkup exclusively in the function

coder2020official (approved):

PR #1580, opened 2022-06-14
Fixed async long polling

e323w:
Hi, looks like that there is a bug in `asyncio_helper.py`, `get_updates()`, due to which `timeout` parameter is not pass into a `getUpdates` function and as a result, short polling is used instead of a long polling.


Badiboy:
It's definitely a mistake. Thank you!

PR #1586, opened 2022-06-23
Added json attribute to main types

kazusman:
## Add self.json attribute to several types in types.py

Hello! I noticed that there is a self.json attribute for types.Message . It seems to me very convenient for logging user requests. For example, I have a bot that performs banking services, and it is important for me to record some user actions. To do this, I take self.json from an instance of the Message class and put it in the database. But other types do not have this attribute, and for example, when I receive CallbackQuery, I can no longer take the original json and write it to the database. It seems to me that it would be more convenient, so I added this attribute to all types that can be received by the bot from the user

## Describe your tests

I created a test bot, to which I sent the necessary types and accessed their json attribute, checking for an exception of type AttributeError

Python version: 3.8.3, 3.9.5, 3.10.5

OS: MacOS 12.4, Linux (Ubuntu 20.04)

## Checklist:
- [X ] I added/edited example on new feature/change (if exists)
- [ ✓] My changes won't break backward compatibility
- [ ✓] I made changes both for sync and async


Badiboy:
Okay, let's assume that it may be useful to record some **significant** user actions.

It's fine for Message, CallbackQuery.

May be for InputTextMessageContent, Poll etc. I do not think so but may be.

But why you added json for classes like ChatMemberUpdated, ChatJoinRequest, ChatPhoto and some other miscelaneous classes/events? It's not seems to be **valuable** user actions.

InlineQueryResultXXX - it's not the one who is sent **from Telegram**. It's sent **from bot**. It even do not have JsonDeserealize. 

ChatLocation is not exists standalone, it's a part of Chat class...

**The set of modified classes is very strange. I cannot see any logic in these modifications.**

kazusman:
## But why you added json for classes like ChatMemberUpdated, ChatJoinRequest, ChatPhoto and some other miscelaneous classes/events? It's not seems to be valuable user actions.

### Answer 

All the types I changed were chosen because they were present in the de_json method of the Update class.

As you can see, I made three commits. The very first one is adding an attribute to the CallbackQuery class. I wanted to stop on it, but then I thought that if I did it for this, then why the rest of the types related to Update will be without this attribute. Therefore, today I made a second commit, in which I added a method to the rest of the classes

I will not take the third commit as an example, I just merged your master into my master


## InlineQueryResultXXX - it's not the one who is sent from Telegram. It's sent from bot. It even do not have JsonDeserealize.

### Answer 

As for InlineQueryResultXXX, I didn’t quite understand what you are talking about. I didn't change anything in those classes. Perhaps you mean ChosenQueryResult, but in this case, this is the event that the bot receives from the user and it does not contradict your words. And the reason for adding the json attribute in this class is the same as in the first paragraph of my answer to the first question


## ChatLocation is not exists standalone, it's a part of Chat class...


### Answer

The answer is the same as for the second question. I didn't change anything in this class

Badiboy:
> As for InlineQueryResultXXX, I didn’t quite understand what you are talking about. I didn't change anything in those classes.

Yep, sorry, my fault.


> As you can see, I made three commits. The very first one is adding an attribute to the CallbackQuery class. I wanted to stop on it, but then I thought that if I did it for this, then why the rest of the types related to Update will be without this attribute. Therefore, today I made a second commit, in which I added a method to the rest of the classes

I do not think that adding functionality without any resons is a good idea. I propose to stop on the first commit that have really valuable reasoning.

Badiboy:
> why not give such access to all events that relate to Update

Because of no reason. "Why not" and "Because we can" is not enough. @coder2020official also likes to make improvements in this way, we always confront about it. :)

Badiboy:
You fix unnecessary part or I close the PR?

Issue #1594, opened 2022-06-29
ERROR - TeleBot: "Unknown error: TimeoutError" (AsyncTeleBot)


chloesimp:
Please answer these questions before submitting your issue. Thanks!

1. What version of pyTelegramBotAPI are you using? 
3.6.0
2. What OS are you using?
Windows 11
3. What version of python are you using?
3.10.5

## What's wrong
I was just running the bot with `asyncio.run(bot.polling(non_stop=True))` then this error came up in the console
```
2022-06-29 23:17:49,349 (asyncio_helper.py:76 MainThread) ERROR - TeleBot: "Unknown error: TimeoutError"
2022-06-29 23:17:49,349 (async_telebot.py:234 MainThread) ERROR - TeleBot: "Request timeout. Request: method=get url=getUpdates params=<aiohttp.formdata.FormData object at 0x000001C1B083B010> files=None request_timeout=20"
```

## Steps to reproduce
1. Run the bot and wait for around 20 seconds. The error will show up. (If the bot doesn't receive any messages for 20 seconds, the error will show up)

coder2020official:
Hello!
I was investigating this issue for a long time. Recently, I have tested an echo-bot example on production server with a good internet connection.
So, the results were good: there were less Timeout errors due to good internet.
You probably have a problem with internet connection, cuz TimeOutError is mostly raised because of internet connection problems. 

My recommendations:
- Increase the number of timeout parameter, for example, from 20 to 40.
- Use non_stop=True parameter. Bot won't crash (I see that you are already using this parameter, it is good)
- Ignore the errors, It shouldn't occur that much on production server

I will continue investigations on this bug.

coder2020official:
And sorry for the late response. Looks like gmailbot is broken, I don't get some notifications from github...

coder2020official:
I made further investigations and it turned out that the module was using short polling by default, which is bad.

coder2020official:
This should be closed

PR #1603, opened 2022-07-02
#1594 & fixes regarding handler execution in async

coder2020official:
#1594 & fixes regarding handler execution in async


coder2020official:
Ready

PR #1631, opened 2022-07-15
update_sensitive field for middlewares

coder2020official:
## Description
- Added update_sensitive field for middlewares. By default, it is false.
If it is set to True, then depending on the [update_type](https://pytba.readthedocs.io/en/latest/types.html#telebot.types.Update), a function will be executed. Here is an example:
image

## Describe your tests
How did you test your change?
I created a basic example for sync and async, then I debugged the functions.

Python version: 3.10.2

OS: Windows

## Checklist:

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -3008,8 +3008,17 @@ class TeleBot:
         :return: None
         """
         if not self.use_class_middlewares:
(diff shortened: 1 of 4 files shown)


coder2020official:
Ready(beta)

PR #1646, opened 2022-07-27
Removed state storages and fixed a typo(for docs)

coder2020official:


coder2020official:
Sorry, I forgot to write that it is ready

PR #1652, opened 2022-08-01
Added InputFile

coder2020official:
## Description
InputFile is an easier way to send files. It is present in API, so I thought it would be useful to have it.
## Describe your tests
How did you test your change?
Tested for all methods(send_xyz)
Python version: 3.10.2

OS: Windows 10

## Checklist:


Badiboy:
Why you closed it?

Badiboy (approved):

PR #1659, opened 2022-08-09
Example of async bot using aiohttp and webhook

Mahakam20000:
## Description
Added example of an async bot using webhook, aiohttp.

## Describe your tests
How did you test your change?

Python version: 3.9

OS: Debian 11.4

## Checklist:


Badiboy:
bot = AsyncTeleBot(API_TOKEN)
auxbot = telebot.TeleBot(API_TOKEN)

Two different bots with one token... What's that?

Mahakam20000:
I'm not skilled in python, it want to be a workaround to use remove_webhook and set_webhook before start the event loop. I knew it was bad. I wrote better code and I'll make a new PR.

PR #1662, opened 2022-08-12
Add 'web_app_data' content type to README

Peibolvig:
Update README to include the valid content type 'web_app_data' for Message objects.

That kind of content type is the one that WebApp bots can send from KeyboardButtons, using the sendData method (More info [here](https://core.telegram.org/bots/webapps#initializing-web-apps))

Debug info from TeleBot with the returned message (edited to avoid leaking real data):

DEBUG - TeleBot: "The server returned: 'b'{"ok":true,"result":[{"update_id":000000000,\n"message":{"m
essage_id":000,"from":{"id":00000000,"is_bot":false,"first_name":"XXX","username":"XXX","language_code":"es"},"chat":{"id":00000000,"first_name":"XXX","use
rname":"XXX","type":"private"},"date":1660260000,"web_app_data":{"button_text":"MyWebAppButton","data":"ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"}}}]}''"


Badiboy:
Thanks.

PR #1688, opened 2022-08-28
Import aioredis from redis module too

ananthb:
## Description
aioredis is available in redis-py as of version 4.2.0rc1: https://github.com/aio-libs/aioredis-py#-aioredis-is-now-in-redis-py-420rc1-
Try importing from the new package as well.

## Describe your tests
How did you test your change?
I tried async storage with aioredis from redis-py.

Python version: 3.10

OS: Linux

## Checklist:

### DIFF
--- a/telebot/asyncio_storage/redis_storage.py
+++ b/telebot/asyncio_storage/redis_storage.py
@@ -6,7 +6,10 @@ redis_installed = True
 try:
     import aioredis
 except:
-    redis_installed = False
+    try:
+        from redis import asyncio as aioredis
+    except:
+        redis_installed = False
 
 
 class StateRedisStorage(StateStorageBase):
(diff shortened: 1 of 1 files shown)


coder2020official:
Never tried. I'll take a look ASAP, thank you

coder2020official:
This version is a pre-release, I think we don't need to replace it now

ananthb:
That's just when it was introduced. Current stable 4.3 and above have the changes already.

coder2020official:
LGTM, thanks.

coder2020official:
But instead of just putting except, it'd be good to make it except ImportError

coder2020official:
Or making 2 try-except blocks 'd look better.. let's wait for badiboy

Badiboy:
I do not see something bad in the proposed method.

coder2020official:
Great! Thanks

PR #1692, opened 2022-08-30
Drop FastAPI for async and use Starlette directly.

ananthb:
## Description
Drop FastAPI for async and use Starlette directly.

Python version: 3.10

OS: Linux

## Checklist:

### DIFF
--- a/examples/asynchronous_telebot/webhooks/run_webhooks.py
+++ b/examples/asynchronous_telebot/webhooks/run_webhooks.py
@@ -37,7 +37,7 @@ async def echo_message(message):
     await bot.reply_to(message, message.text)
 
 
-# it uses fastapi + uvicorn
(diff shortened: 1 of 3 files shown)


coder2020official:
Not reasonable.

coder2020official:
image
image
Are you talking about these things? I hope they will update the dependencies soon. Using fastapi directly is not a bad idea.

ananthb:
I mean that replacing FastAPI with Starlette directly is backwards compatible with the code running today.

This only uses the portion of FastAPI that is directly taken from Starlette, so why not just use Starlette directly? My code looks and works largely like the FastAPI code without lesser dependencies.

coder2020official:
It isn't a good reason

ananthb:
I understand. Thanks!

PR #1694, opened 2022-08-31
request_timeout field for send_media_group

kuziukov:
## Description
The problem I found is not possible to set Timeout for 'send_media_group', so there is a solution.
The method 'send_media_group' do not use 'request_timeout' of _process_request, it always stays as None.

Python version: 3.10

OS: Big Sur

## Checklist:

### DIFF
--- a/telebot/async_telebot.py
+++ b/telebot/async_telebot.py
@@ -3065,7 +3065,8 @@ class AsyncTeleBot:
             protect_content: Optional[bool]=None,
             reply_to_message_id: Optional[int]=None, 
             timeout: Optional[int]=None,
-            allow_sending_without_reply: Optional[bool]=None) -> List[types.Message]:
+            allow_sending_without_reply: Optional[bool]=None,
+            request_timeout: Optional[int]=None) -> List[types.Message]:
         """
         Use this method to send a group of photos, videos, documents or audios as an album. Documents and audio files
         can be only grouped in an album with messages of the same type. On success, an array of Messages that were sent is returned.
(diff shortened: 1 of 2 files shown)


kuziukov:
Look, I use async Telebot and I cannot set timeout for large media requests because it is always None. In the function `_process_request` you gays uses 'request_timeout' but it is not possible to set from `send_media_group`.

Badiboy:
```
async def _process_request(token, url, method='get', params=None, files=None, request_timeout=None):
    params = _prepare_data(params, files)
    if request_timeout is None:
        request_timeout = REQUEST_TIMEOUT
    timeout = aiohttp.ClientTimeout(total=request_timeout)
```

@coder2020official Seems you should get request_timeout from params["timeout"] (if exists) if request_timeout is None. Because all sending functions cannot pass request_timeout to _process_request.

Something like this:

```
    if request_timeout is None:
        if "timeout" in params:
                request_timeout = params["timeout"]
        else:
                request_timeout = REQUEST_TIMEOUT

```

kuziukov:
Yep, I guess your solution is good

kuziukov:
Or you can implement something like this in `asyncio_helper.py`.

```
async def send_media_group(
        token, chat_id, media,
        disable_notification=None, reply_to_message_id=None,
        timeout=None, allow_sending_without_reply=None, protect_content=None):
    method_url = r'sendMediaGroup'
    media_json, files = await convert_input_media_array(media)
    payload = {'chat_id': chat_id, 'media': media_json}
    if disable_notification is not None:
        payload['disable_notification'] = disable_notification
    if reply_to_message_id:
        payload['reply_to_message_id'] = reply_to_message_id
    if timeout:
        payload['timeout'] = timeout
    if allow_sending_without_reply is not None:
        payload['allow_sending_without_reply'] = allow_sending_without_reply
    if protect_content is not None:
        payload['protect_content'] = protect_content
    return await _process_request(
        token, method_url, params=payload,
        method='post' if files else 'get',
        files=files if files else None,
        request_timeout=timeout
    )
```

coder2020official:
I think this should be moved to #1695 
request_timeout -> is a temporary field only for get_updates.
timeout is basically the same thing with request_timeout, but this is different for get_updates method, cuz the method itself accepts timeout parameter.

PR #1695, opened 2022-09-01
Fix timeout usage in AsyncTeleBot

Badiboy:
## Description
_timeout_ parameter of API functions was ignored in Async version.

Alternative fix for request: https://github.com/eternnoir/pyTelegramBotAPI/pull/1694

### DIFF
--- a/telebot/asyncio_helper.py
+++ b/telebot/asyncio_helper.py
@@ -60,7 +60,7 @@ async def _process_request(token, url, method='get', params=None, files=None, re
     params = _prepare_data(params, files)
     if request_timeout is None:
         if params and ("timeout" in params):
-            request_timeout = params["timeout"]
+            request_timeout = params.pop("timeout")
         else:
             request_timeout = REQUEST_TIMEOUT
     timeout = aiohttp.ClientTimeout(total=request_timeout)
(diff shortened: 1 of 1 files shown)


telebot/asyncio_helper.py
```diff
@@ -59,7 +59,10 @@ async def get_session(self):
 async def _process_request(token, url, method='get', params=None, files=None, request_timeout=None):
     params = _prepare_data(params, files)
     if request_timeout is None:
-        request_timeout = REQUEST_TIMEOUT
+        if params and ("timeout" in params):
```
coder2020official, line 62: As you can see [here](https://github.com/eternnoir/pyTelegramBotAPI/blob/master/telebot/async_telebot.py#L146), get_updates has 2 params: request_timeout & timeout. Timeout should be passed to telegram, meanwhile request_timeout is supposed to go to aiohttp. So, if the method is not getUpdates, then it should pop the timeout from dictionary. For getUpdates, request_timeout and timeout should be passed without changes.

Badiboy: > So, if the method is not getUpdates, then it should pop the timeout from dictionary

@coder2020official Please, check.

coder2020official:
Moved to #1710 ?

PR #1715, opened 2022-09-12
Fix incorrect webhook URL port in aiohttp examples

Y-T-G:
## Description
Updated examples to use default port in webhook URL. Resolves issues such as [this](https://stackoverflow.com/questions/65582845/use-an-open-port-when-deploying-telegram-bot-using-pytelegrambotapi-to-heroku/73693670).

## Describe your tests
Tested by deploying on Heroku with the above change. Webhook running fine.

Python version: 3.10.7

OS: Ubuntu 20.04

## Checklist:

### DIFF
--- a/examples/asynchronous_telebot/webhooks/async_webhook_aiohttp_echo_bot.py
+++ b/examples/asynchronous_telebot/webhooks/async_webhook_aiohttp_echo_bot.py
@@ -17,7 +17,7 @@ WEBHOOK_PORT = 8443  # 443, 80, 88 or 8443 (port need to be 'open')
 WEBHOOK_LISTEN = '0.0.0.0'  # In some VPS you may need to put here the IP addr
 WEBHOOK_SSL_CERT = './webhook_cert.pem'  # Path to the ssl certificate
 WEBHOOK_SSL_PRIV = './webhook_pkey.pem'  # Path to the ssl private key
-WEBHOOK_URL_BASE = "https://{}:{}".format(WEBHOOK_HOST, WEBHOOK_PORT)
+WEBHOOK_URL_BASE = "https://{}".format(WEBHOOK_HOST)
 WEBHOOK_URL_PATH = "/{}/".format(API_TOKEN)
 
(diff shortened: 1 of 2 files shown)


Badiboy:
Why you decided that if you cannot deploy this example to Heroku (one of quite specific services) - the example is incorrect and should be updated?

If you want to have specific Heroku example - make one. Or add HEROKU define for ones who want to use it on Heroku...

Y-T-G:
@Badiboy Why would one need to specify a port in the webhook URL that is going to be used by Telegram API? The only port that needs to be specified is the port that the application is going to be listening to.

Badiboy:
@Y-T-G Because you may have multiple bots or other services on one IP.

coder2020official:
Those who need it will edit example under their needs;

Y-T-G:
@Badiboy @coder2020official Okay, how about I separate external and internal port in two different variables,, since the external port can only be 80, 443, 88 and 8443, while the internal port that webserver listens to can be anything?

Badiboy:
> since the external port can only be 80, 443, 88 and 8443

Where did you get it?

coder2020official:
image
Ig he is talking about this thing.
Btw, this is not a MISTAKE. Anyone who does not need port - can edit example under their needs; Not on the official repo;

Badiboy:
> Where did you get it

@coder2020official You are right. Sorry, I forget that I manage channels with different hook URL, not port.

byehack:
@Y-T-G you can use this for heroku:
```py
int(os.environ.get('PORT', 5000))
```

Y-T-G:
@coder2020official It's not really clear from the examples that one could use a different internal port since there's no differentiation being done between the external and internal ports. A single define sets both the internal and external ports.

coder2020official:
Other examples have it as well. No need to edit examples: examples show basic usage. For further needs, user should edit the example under his/her needs.

Badiboy:
Enough, sorry.

PR #1706, opened 2022-09-09
Accept handlers return closes #1704

byehack:
basic implement for https://github.com/eternnoir/pyTelegramBotAPI/pull/1704

- [ ] Test
- [ ] Create example
- [ ] Document
- [ ] Bump new version

Issue #1704, opened 2022-09-08
improve multi handles on one message.


byehack:
Currently nothing do on handler functions return.
but we can use to improve handle matching on function returns:
see this example:

```python3

from telebot.async_telebot import AsyncTeleBot
from telebot.return_types import ContinueHandling # new return_types

bot = AsyncTeleBot('TOKEN')
database = ...


@bot.message_handler(commands=['help'])
async def cmdlist(message):
    if message.chat.id not in database:
        await bot.reply_to(message, "Hello, I see you first time.")
    return ContinueHandling


@bot.message_handler(commands=['help'])
async def cmdlist(message):
    await bot.reply_to(message, "helpmsg")
```

byehack:
continue discussion in https://github.com/eternnoir/pyTelegramBotAPI/pull/1706

PR #1722, opened 2022-09-17
Handlers and Middlewares processing union

Badiboy:
## Description
Call for handlers now union in a single function for future extension.
Plus minor fixes in storages.

## Checklist:

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -5876,7 +5876,7 @@ class TeleBot:
             return False
 
     # middleware check-up method
-    def _check_middlewares(self, update_type):
+    def _get_middlewares(self, update_type):
         """
         Check middleware
 
@@ -5988,7 +5988,7 @@ class TeleBot:
             return
 
         if self.use_class_middlewares:
-            middlewares = self._check_middlewares(update_type)
+            middlewares = self._get_middlewares(update_type)
         else:
             middlewares = None
         for message in new_messages:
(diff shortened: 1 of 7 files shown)


telebot/__init__.py
```diff
@@ -5875,7 +5876,7 @@ def _check_filter(self, message_filter, filter_value, message):
             return False
 
     # middleware check-up method
-    def _check_middleware(self, update_type):
+    def _check_middlewares(self, update_type):
```
coder2020official, line 5879: I think we can rename this function to _get_middlewares

telebot/__init__.py
```diff
@@ -5889,100 +5890,111 @@ def _check_middleware(self, update_type):
 
     def _run_middlewares_and_handler(self, message, handlers, middlewares, update_type):
```
coder2020official, line 5891: And this function should be renamed to smth like _run_update_task

coder2020official (changes requested):
~ I will test it tonight.

coder2020official:
- This pull request will:
- help to #1704 (by making it possible to get the result of a handler without middlewares;
- Improve the readability of code

telebot/__init__.py
Badiboy: +

telebot/__init__.py
Badiboy: Don't like this name. Let's left it for next updates.

byehack:
@Badiboy @coder2020official 
I think by #1706 we can deprecate middleware functions, and remove it in next major version.
You can open a draft PR to start deprecating middleware part.
if you still know a usecase of middlewares after merging #1706 , I would happy to know.

coder2020official:
Middleware functions are another way of using middlewares. We are not going to deprecate it for now

PR #1739, opened 2022-09-28
update for test environments

xsa-dev:
## Description
Added test in apihelper.

## Describe your tests
How did you test your change?
I use test telegram environment for developing telegram bots

Python version:
3.10

OS:
Mac OS X

## Checklist:
- [+] I added/edited example on new feature/change (if exists)
- [+] My changes won't break backward compatibility
- [+] I made changes both for sync and async


Badiboy:
What's that?!

Just set API_URL in yur code to what you need. It is presented especially for customizing URLs.

xsa-dev:
Thanks!

PR #1773, opened 2022-10-29
Check if the event loop is running (async telebot)

David256:
## Description
Check if the event loop is running, otherwise create a new event loop.

## Describe your tests
The changes were tested on a bot of our own because this library does not yet have tests for async.

Python version: Python 3.9.2

OS: Debian GNU/Linux 11

### DIFF
--- a/telebot/async_telebot.py
+++ b/telebot/async_telebot.py
@@ -174,6 +174,10 @@ class AsyncTeleBot:
         self.middlewares = []
 
         self._user = None # set during polling
+        try:
+            asyncio.get_running_loop()
+        except RuntimeError as e:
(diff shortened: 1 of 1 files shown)


coder2020official:
Why?

David256:
Usually raises `Event loop is closed` the first time when AsyncTeleBot sends `getUpdates`.

coder2020official:
This was a windows bug; There are no such issues on other platforms;
And, setting new event loop on object creation is a very bad idea.

David256:
Well, simply calling `asyncio.get_event_loop()` has the same effect because it checks for an event loop. But, this was the solution I found for this. 🤔

Issue #1767, opened 2022-10-27
`debug` removed from uvicorn


batmanscode:
Change in uvicorn: [Remove --debug flag #1640](https://github.com/encode/uvicorn/pull/1640)

---

`run_app()` https://github.com/eternnoir/pyTelegramBotAPI/blob/ebec3bf5c11cd685e1f3f031a422cfd831f7f428/telebot/ext/sync/webhooks.py#L107 and `run_webhooks()` https://github.com/eternnoir/pyTelegramBotAPI/blob/ebec3bf5c11cd685e1f3f031a422cfd831f7f428/telebot/__init__.py#L441 will need to be updated.

coder2020official:
Can you do a PR?

batmanscode:
Yes @coder2020official

PR #1774, opened 2022-10-30
fixes: `debug` removed from uvicorn #1767

batmanscode:
## Description
`--debug` was removed from uvicorn. See #1767.

## Describe your tests
Untested; Just removed debug as a param since that was causing `TypeError: run() got an unexpected keyword argument 'debug'` when using `run_webhooks()`.

Python version: 3.9

## Checklist:


Badiboy:
@coder2020official Looks fine.

PR #1816, opened 2022-11-29
Translated into russian file: formatting.po

Cub11k:
## Description

## Describe your tests
How did you test your change?

Python version:

OS:

## Checklist:

### DIFF
--- a/docs/source/locales/ru/LC_MESSAGES/formatting.po
+++ b/docs/source/locales/ru/LC_MESSAGES/formatting.po
@@ -20,23 +20,23 @@ msgstr ""
 
 #: ../../formatting.rst:3
 msgid "Formatting options"
-msgstr ""
+msgstr "Параметры форматирования"
 
 #: ../../formatting.rst:5
 msgid "Formatting options in pyTelegramBotAPI"
-msgstr ""
+msgstr "Параметры форматирования в pyTelegramBotAPI"
 
 #: ../../formatting.rst:5
 msgid "html, markdown, parse_mode, formatting, ptba, pytba, pyTelegramBotAPI"
-msgstr ""
+msgstr "html, markdown, parse_mode, форматирование, ptba, pytba, pyTelergamBotAPI"
 
 #: of telebot.formatting:1
 msgid "Markdown & HTML formatting functions."
-msgstr ""
+msgstr "Функции форматирования Markdown & HTML."
 
 #: of telebot.formatting.escape_html:1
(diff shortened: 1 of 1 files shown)


docs/source/locales/ru/LC_MESSAGES/formatting.po
```diff
@@ -20,23 +20,23 @@ msgstr ""
...
 
 #: of telebot.formatting:1
 msgid "Markdown & HTML formatting functions."
-msgstr ""
+msgstr "Функции форматирования Markdown & HTML."
 
 #: of telebot.formatting.escape_html:1
 msgid "Escapes HTML characters in a string of HTML."
-msgstr ""
+msgstr "Экранирует HTML символы в HTML строке."
```
coder2020official, line 39: this doesn't sound good.. what about making it "пропускает" ?

docs/source/locales/ru/LC_MESSAGES/formatting.po
```diff
@@ -147,105 +147,105 @@ msgstr ""
 #: telebot.formatting.mspoiler:6 telebot.formatting.mstrikethrough:6
 #: telebot.formatting.munderline:6
 msgid "True if you need to escape special characters. Defaults to True."
-msgstr ""
+msgstr "True если вам нужно экранировать спец. символы. По умолчанию True."
```
coder2020official, line 150: you can use "избегать" here i think

docs/source/locales/ru/LC_MESSAGES/formatting.po
Cub11k: Agreed. it doesn't sound good, but the thing is that this function doesn't really skip("пропускает") HTML characters. It replaces them with their codes, making them non-HTML for parser.

docs/source/locales/ru/LC_MESSAGES/formatting.po
Cub11k: Same as my respond for previous comment. It doesn't really skip or avoid characters, it does replace them with their codes, making them non-HTML for parser.

docs/source/locales/ru/LC_MESSAGES/formatting.po
Cub11k: Might be convinient enough to use "пропускает" cause parser will skip escaped characters. Will make an update

coder2020official (approved):

coder2020official:
LGTM, thanks

PR #1824, opened 2022-12-03
fixing escape()

reddere:
fixing escape() as replacing a None would throw an exception 'NoneType' object has no attribute 'replace'. useful in case of escaping a None string given from message.from_user.last_name as you dont know wether the user has a last name or not

### DIFF
--- a/telebot/util.py
+++ b/telebot/util.py
@@ -401,8 +401,7 @@ def escape(text: str) -> str:
     chars = {"&": "&amp;", "<": "&lt;", ">": "&gt;"}
     if text is None:
         return None
-    else:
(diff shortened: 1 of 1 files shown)


Badiboy:
Suitable update.

Please:
`if text == None:`
Bad practice. Replace to
`if text is None:`

And "else" after that not needed.

reddere:
thanks! changed the "== None" to "is None", what do you mean by "And "else" after that not needed."?

reddere:
should be fine now, let me know if any more changes are needed!

Badiboy:
👍

PR #1839, opened 2022-12-19
Fix bug with asyncwebhooks

coder2020official:


Badiboy:
image

as far as AsyncWebhookListener have named parameters:

image

I propose to pass them as named when call constructor. That will also allow to avoid such bugs as it occurs with "debug".

Badiboy (approved):
👍

PR #1844, opened 2022-12-20
Fix #1843

### LINKED ISSUE #1843: Incorrect behaviour of escape_markdown
Please answer these questions before submitting your issue. Thanks!

1. What version of pyTelegramBotAPI are you using?
4.8.0
2. What OS are you using?
Both Linux/Windows 10
3. What version of python are you using?
Python 3.9.0

Function `escape_markdown` from `telebot.formatting` doesn't escape symbols `{` and `}`, however according to the [core.telegram](https://core.telegram.org/bots/api#markdownv2-style) they must be escaped with the preceding character `\`.

It means that now sending messages using `parse_mode='MarkdownV2'` gives the following error when the message text contains symbols `{` and `}`: `A request to the Telegram API was unsuccessful. Error code: 400. Description: Bad Request: can't parse entities: Character '{' is reserved and must be escaped with the preceding '\'`.

coder2020official:
#1843

### DIFF
--- a/telebot/formatting.py
+++ b/telebot/formatting.py
@@ -61,8 +61,8 @@ def escape_markdown(content: str) -> str:
     :rtype: :obj:`str`
     """
     
-    parse = re.sub(r"([_*\[\]()~`>\#\+\-=|\.!])", r"\\\1", content)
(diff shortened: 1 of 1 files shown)


Badiboy (commented):
Up to you. I'm bad in regexp. Sorry.

PR #1852, opened 2022-12-28
Fixed typehints for register_<any>_handler()

Cub11k:
## Description
Fixed super annoying incorrect typehints in register_<any>_handler() methods of AsyncTeleBot
Replaced Awaitable by Callable[[Any], Awaitable]

## Describe your tests
Defined asynchronous functions with different number of arguments (Any) and different return types
Tried to register them as a handler for AsyncTeleBot - got no warnings for typehints

Python version:

OS:

## Checklist:


coder2020official (approved):
LGTM, thanls

PR #1854, opened 2022-12-30
Fixed typehints for reply markup in editing methods

Cub11k:
## Description
Editing methods allow to pass only inline keyboards as reply markup. Fixed typehints accordingly

## Describe your tests
How did you test your change?

Python version:

OS:

## Checklist:


Badiboy:
Thank you!

PR #1856, opened 2022-12-30
Copied translations from sync_version.po to async_version.po

Cub11k:
### Extra:
Minor fixes in sync_version.po, found while copying

## Describe your tests
How did you test your change?

Python version:

OS:

## Checklist:


coder2020official (approved):
LGTM, thanks

PR #1867, opened 2023-01-05
Translated into russian files: index.po, calldata.po

Cub11k:
## Description

## Describe your tests
How did you test your change?

Python version:

OS:

## Checklist:


coder2020official:
Thank you

Issue #1864, opened 2023-01-04
Circular import in types & util files


coder2020official:
Types file imports util:
https://github.com/eternnoir/pyTelegramBotAPI/blob/master/telebot/types.py#L15
And util imports types:
https://github.com/eternnoir/pyTelegramBotAPI/blob/master/telebot/util.py#L16

@Badiboy any ideas on properly fixing the issue?

Badiboy:
There were no any **problems** with circular import before.
When the **problem** appears?

coder2020official:
I've seen some reports before in chats, but they fixed it somehow.
Here's another screenshot of bugreport i got today:
![telegram-cloud-photo-size-2-5307926708923515092-y](https://user-images.githubusercontent.com/63740404/210575743-0f6d4564-ab74-44f7-b2ee-7374df2abeb6.jpg)
And the links in the issue description show the circular import. When one of the files import another file, it imports the file that it importing it. It is definitely a circular import issue

Badiboy:
Thanks, I'll check.

Badiboy:
https://github.com/eternnoir/pyTelegramBotAPI/pull/1868

PR #1868, opened 2023-01-06
Fix circular import

Cub11k:
## Description
Fix for issue #1864
Added __all__ to util.py - for sphinx to generate docs accordingly

Extras:
Update README - add ru docs link

## Describe your tests
How did you test your change?

Python version:

OS:

## Checklist:


Cub11k:
I see that my pep8 formatter made some changes, this wasn't intentional, sorry xD

coder2020official:
Please next time focus on one thing per PR.
Thoughts @Badiboy ?

Badiboy:
Elegant enough except unit name :) I propose something like "library_util" or "service_functions"...

Cub11k:
@Badiboy maybe "_service" looks good?

Badiboy:
@Cub11k For me personally, I do not like names like _util or _service... I prefer full qualified names, but it's very subjective. If you and @coder2020official like "_util" or "_service" - let it be. It was just my personal oppinion for naming.

Cub11k:
Renamed to "service_utils", looks fine for me

Badiboy:
@Cub11k  👌

The last thing:

image

What for?

Cub11k:
@Badiboy It's for sphinx to generate docs for methods, that are in service_utils.py now

Tried without _ _all_ _ - didn't work, maybe it requires some change it util.rst, but I'm not very familiar with it's behaviour

Badiboy:
Let it be :)

telebot/service_utils.py
```diff
@@ -0,0 +1,85 @@
+import random
+import string
+
+try:
+    # noinspection PyPackageRequirements
+    from PIL import Image
+    from io import BytesIO
+
+    pil_imported = True
+except:
```
Tishka17, line 10: It should be `except ImportError` ?

telebot/service_utils.py
```diff
@@ -0,0 +1,85 @@
+import random
+import string
+
+try:
+    # noinspection PyPackageRequirements
+    from PIL import Image
+    from io import BytesIO
```
Tishka17, line 7: No ideas why problems wirh this import are expected and threaten as no PIL installed.

Probably this import should be done outside of try block

Cub11k: PIL is not enlisted in package requirements, so it's the same as with redis for StateRedisStorage

telebot/service_utils.py
Cub11k: Agree, it definitely should be
That's why ctrl+C ctrl+V is bad xD

PR #1873, opened 2023-01-08
Little code style improvement in service_utils

coder2020official:
## Description
- Added ImportError to block except to be more specific on errors related to pillow lib
- Removed io lib from try/except block since it is built-in lib

## Describe your tests
How did you test your change?

Python version:

OS:

## Checklist:

### DIFF
--- a/telebot/service_utils.py
+++ b/telebot/service_utils.py
@@ -1,11 +1,10 @@
 import random
 import string
+from io import BytesIO
 
 try:
     # noinspection PyPackageRequirements
     from PIL import Image
-    from io import BytesIO
-
     pil_imported = True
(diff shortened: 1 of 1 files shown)


coder2020official:
@Tishka17

coder2020official:
@Badiboy do you think it is worth releasing a new version with hotfix?

Badiboy:
This cross-reference exists for months and appears in a very rare cases, so I do not think release of this update is urgent.

However, release an update is not a problem, if you think it's needed.

Cub11k:
@coder2020official Same thing in StateRedisStorage, too broad except on redis import🕺

PR #1881, opened 2023-01-10
Create method get_media_file_id() in util.py

Cub11k:
## Description
This method is used to get file id of different types of media

## Describe your tests
How did you test your change?

Python version:

OS:

## Checklist:

### DIFF
--- a/telebot/util.py
+++ b/telebot/util.py
@@ -643,6 +643,40 @@ def validate_web_app_data(token: str, raw_init_data: str):
     return hmac.new(secret_key.digest(), data_check_string.encode(), sha256).hexdigest() == init_data_hash
 
 
+def get_media_file_id(message: types.Message, content_types: Optional[List[str]] = None):
+    """
+    Use this method to get file id of message media if it's present.
+
+    :param message: The message to get media from
+    :type message: :class:`telebot.types.Message`
+
+    :param content_types: Types of media to look for, by default all types are considered.
+    :type content_types: :obj:`list[str]`, optional
+
+    :return: file_id or :obj:`None`
+    :rtype: :obj:`str` | :obj:`None`
+    """
+    if content_types is None:
+        content_types = content_type_media
+
+    if "animation" in content_types and message.animation is not None:
+        return message.animation.file_id
+    elif "audio" in content_types and message.audio is not None:
+        return message.audio.file_id
+    elif "document" in content_types and message.document is not None:
+        return message.document.file_id
+    elif "photo" in content_types and message.photo is not None:
+        return message.photo[0].file_id
+    elif "sticker" in content_types and message.sticker is not None:
+        return message.sticker.file_id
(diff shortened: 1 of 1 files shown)


coder2020official:
image

I think it would be better to make a list with all content types supported; then use getattr() directly. this would reduce this mess...

coder2020official:
@Badiboy next time please don't hurry with merging

Badiboy:
@coder2020official OK. 

It's functional. If anyone want to optimize it - it will not affect the result. So it can be merged. But OK: I got you.

coder2020official:
@Cub11k will you re-create PR or should I do it myself?

Cub11k:
@coder2020official Sure thing, you can re-create PR

coder2020official:
@Badiboy ?

Badiboy:
I already said my thoughts.

> It's functional. If anyone want to optimize it - it will not affect the result. So it can be merged.

I also think that there is too much attention to trivial update. If you want to optimize it - just do it.

coder2020official:
Read my previous comment. i don't think it should be in the lib.

Badiboy:
I see no problem of being nearly ANY function in the lib. If someone think it is useful (I also think it may be) - why not.

coder2020official:
BRUH
Then why don't we add any other simple functions, such as a function to do if-else?
This function is useless, since the thing can be quickly done through getattr and for loop.
Your thoughts have changed A LOT since my first talk to you - you asked for reason each time i was adding smth new that time. Then why don't u ask for a reason before merging this PR? Why don't u at least wait for my thought?

Badiboy:
Looks like you have a lot of free time to discuss this update...

> you asked for reason each time i was adding smth new that time.

You tried to do it in the MAIN ENGINE. In the main engine useless or very rarely used things affects the performance of all millions of systems build on this library. Functions in util do not affect anything until you call them.

More then a year had passed. You still do not understand the difference between main engine and utils. That's sad.

coder2020official:
Let me remind you: my functions were not mostly based on MAIN ENGINE, as you said. They were mostly functions in main class. So, technically they are the same as functions in util - they don't mess up until you call them.
And you asked for a reason each time. Even when i was adding features like custom filters / states. That seemed to be ridiculous, however, now you need to do that before merging.

The sad thing is that you no more ask for reasons and don't wait for my opinion. Don't make your resentment or personal opinion ruin the project.

coder2020official:
So, I'd better ask to remove this function since it is useless. Your opinion?

Note that i never merge PRs unless it is docs-related or some minor improvement is made. I've always asked for your opinion and I 'd ask you to do the same.

Badiboy:
I can not waste any more time to repeat the same thing I said already 2 times. Just read above.

coder2020official:
Ok then I'll make PR with deletion of this func

Badiboy:
No, not OK. Read again.

PR #1884, opened 2023-01-16
Remove redundant function from util

Cub11k:
## Description
This function can be replaced with this simple code:

```
file_id = None
content_types = ["photo", "video", ...]
for content_type in content_types:
    media = getattr(message, content_type)
    if media is not None:
        file_id = media.file_id
        break
```

Thus it can be removed

## Describe your tests
How did you test your change?

Python version:

OS:

## Checklist:

### DIFF
--- a/telebot/util.py
+++ b/telebot/util.py
@@ -643,40 +643,6 @@ def validate_web_app_data(token: str, raw_init_data: str):
     return hmac.new(secret_key.digest(), data_check_string.encode(), sha256).hexdigest() == init_data_hash
 
 
-def get_media_file_id(message: types.Message, content_types: Optional[List[str]] = None):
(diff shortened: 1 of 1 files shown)


Cub11k:
I hope this would resolve your discussion in #1881 
@coder2020official @Badiboy

coder2020official (approved):

Badiboy:
image

coder2020official:
I'll approve this, but I don't think badiboy is gonna like that.

coder2020official:
@Badiboy you can't just add random function to utils, just because it has no effect on main class.

Badiboy:
@coder2020official Госсподи, какой же ты душнила...

PR #1894, opened 2023-01-28
Added poll_example.py in the examples

Muhammad-Aadil:
## Description
An example file on how to create polls and handle poll answers

Python version:  3.9.7

OS: Ubuntu 20.04.1

### DIFF
--- a/examples/poll_example.py
+++ b/examples/poll_example.py
@@ -1,6 +1,6 @@
 #!/usr/bin/python
 
-# This is an example file to create polls and handle poll answers
+# This is an example file to create quiz polls
 import telebot
 
 API_TOKEN = "<api_token>"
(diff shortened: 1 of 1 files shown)


coder2020official:
I'd recommend making quiz type poll instead

Muhammad-Aadil:
You mean something like this:
`    poll = bot.send_poll(
        chat_id=message.chat.id,
        question="We are going to '' park.",
        options=answer_options,
        **type='quiz',
        correct_option_id=2,**
        is_anonymous=False,
    )`

coder2020official:
Yes. I think this would be better, no?

Muhammad-Aadil:
Yes, it is better, How about we add both?

coder2020official:
I don't think both will look good, since for quizes quiz mode should be used.

Muhammad-Aadil:
Hi! Yes, you are absolutely right.

coder2020official (approved):
Looks fine, any ideas or review @Badiboy ?

coder2020official:
@badiboy can u merge this and look at my bot api pr? since u have unsubscribed from notifications i was unsure whether u were getting an email about me tagging u there.

PR #1902, opened 2023-02-04
Bot API 6.5 update 🔥

coder2020official:
February 3, 2023
Bot API 6.5

- [x] Added [requests for users and chats](https://telegram.org/blog/profile-pics-emoji-translations#chat-selection-for-bots) and support for [granular media permissions](https://telegram.org/blog/profile-pics-emoji-translations#granular-media-permissions).
- [x] Added the class [KeyboardButtonRequestUser](https://core.telegram.org/bots/api#keyboardbuttonrequestuser) and the field request_user to the class [KeyboardButton](https://core.telegram.org/bots/api#keyboardbutton).
- [x] Added the class [KeyboardButtonRequestChat](https://core.telegram.org/bots/api#keyboardbuttonrequestchat) and the field request_chat to the class [KeyboardButton](https://core.telegram.org/bots/api#keyboardbutton).
- [x] Added the classes [UserShared](https://core.telegram.org/bots/api#usershared), [ChatShared](https://core.telegram.org/bots/api#chatshared) and the fields user_shared, and chat_shared to the class [Message](https://core.telegram.org/bots/api#message).
- [x] Replaced the fields can_send_media_messages in the classes [ChatMemberRestricted](https://core.telegram.org/bots/api#chatmemberrestricted) and [ChatPermissions](https://core.telegram.org/bots/api#chatpermissions) with separate fields can_send_audios, can_send_documents, can_send_photos, can_send_videos, can_send_video_notes, and can_send_voice_notes for different media types.
- [x] Added the parameter use_independent_chat_permissions to the methods [restrictChatMember](https://core.telegram.org/bots/api#restrictchatmember) and [setChatPermissions](https://core.telegram.org/bots/api#setchatpermissions).
- [x] Added the field user_chat_id to the class [ChatJoinRequest](https://core.telegram.org/bots/api#chatjoinrequest).

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -2928,8 +2928,7 @@ class TeleBot:
         Telegram documentation: https://core.telegram.org/bots/api#restrictchatmember
 
         .. warning::
-            Please pass `telebot.types.ChatPermissions` object to `permissions` parameter instead of
-            passing all boolean parameters. Those boolean parameters won't be supported soon, so please take it into consideration.
+            Individual parameters are deprecated and will be removed, use 'permissions' instead.
 
         :param chat_id: Unique identifier for the target group or username of the target supergroup
             or channel (in the format @channelusername)
@@ -2997,8 +2996,7 @@ class TeleBot:
                 can_pin_messages=can_pin_messages
             )
             logger.warning(
-                'Please pass `telebot.types.ChatPermissions` object to `permissions` parameter instead of '
-                'passing all boolean parameters. Those boolean parameters won\'t be supported soon, so please take it into consideration.'
+                "Individual parameters are deprecated and will be removed, use 'permissions' instead."
             )
         return apihelper.restrict_chat_member(
             self.token, chat_id, user_id, permissions, until_date, use_independent_chat_permissions)
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -3138,7 +3138,8 @@ class TeleBot:
         return apihelper.unban_chat_sender_chat(self.token, chat_id, sender_chat_id)
 
     def set_chat_permissions(
-            self, chat_id: Union[int, str], permissions: types.ChatPermissions) -> bool:
+            self, chat_id: Union[int, str], permissions: types.ChatPermissions,
+            use_independent_chat_permissions: Optional[bool]=None) -> bool:
         """
         Use this method to set default chat permissions for all members.
         The bot must be an administrator in the group or a supergroup for this to work
@@ -3153,10 +3154,16 @@ class TeleBot:
         :param permissions: New default chat permissions
         :type permissions: :class:`telebot.types..ChatPermissions`
 
+        :param use_independent_chat_permissions: Pass True if chat permissions are set independently. Otherwise,
+            the can_send_other_messages and can_add_web_page_previews permissions will imply the can_send_messages,
+            can_send_audios, can_send_documents, can_send_photos, can_send_videos, can_send_video_notes, and
+            can_send_voice_notes permissions; the can_send_polls permission will imply the can_send_messages permission.
+        :type use_independent_chat_permissions: :obj:`bool`
+
         :return: True on success
         :rtype: :obj:`bool`
         """
-        return apihelper.set_chat_permissions(self.token, chat_id, permissions)
+        return apihelper.set_chat_permissions(self.token, chat_id, permissions, use_independent_chat_permissions)
 
     def create_chat_invite_link(
             self, chat_id: Union[int, str],
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -2918,6 +2918,7 @@ class TeleBot:
             can_change_info: Optional[bool]=None,
             can_invite_users: Optional[bool]=None, 
             can_pin_messages: Optional[bool]=None,
+            permissions: Optional[types.ChatPermissions]=None,
             use_independent_chat_permissions: Optional[bool]=None) -> bool:
         """
         Use this method to restrict a user in a supergroup.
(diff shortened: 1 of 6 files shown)


coder2020official:
In the commit above i just moved class to another place to fix the issue, so no need for review for this commit above

coder2020official:
Commit above should stop issues related to definition time, read [pep-563 ](https://peps.python.org/pep-0563/)for more

Badiboy:
@coder2020official Unsibscribed due to lot of notifications. Pin me when will be ready for review or need something from me.

coder2020official:
Sure

coder2020official:
Oh shit.... @badiboy, our restrictchatmember method is different from docs, we have all permissions in the method, meanwhile the doc has just one parameter permissions used to pass all permissions... how do i add use_independent_chat_permissions?
Doc: https://core.telegram.org/bots/api#restrictchatmember
Method:
image

Badiboy:
@coder2020official We can add optional parameter **permissions** so both old-style and new-style call mode can be used: if "permissions" passed - it has the priority. If no - it is combined from other parameters + deprecation warning.

coder2020official:
ah wait... ig we can still continue passing use_independent_chat_permissions with all permissions for now, and you can fix it later after this PR;

Badiboy:
We can :)

coder2020official:
Or maybe after this pull request i can go through all methods and fix them by keeping backward compatibility; And, there are some methods where we can delete backward-compatible things

coder2020official:
I'm done, @Badiboy ; But i'm unsure about changes, can u go through?

Badiboy:
I will

telebot/types.py
```diff
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+from __future__ import annotations
```
Badiboy, line 2: What's for?

Badiboy:
I'm done. Everithing is fine except 3 points.

telebot/types.py
coder2020official: This is described in pep [563](https://peps.python.org/pep-0563/):
_This PEP proposes changing function annotations and variable annotations so that they are no longer evaluated at function definition time. Instead, they are preserved in __annotations__ in string form._

So, we can define class at any point of file and use it at any point; Basically, you don't have to define a class before using it - you can define it later, after using it.

coder2020official:
Which 3 points? I only see 1 point

telebot/types.py
coder2020official: This was useful for using chatadministratorrights class before it's definition type, and i'm sure it is gonna be useful in future

Badiboy:
1.
image

2.
image

coder2020official:
ok, u probably forgot to send these changes

coder2020official:
btw, i forgot to add these changes to restrict chat member

coder2020official:
Btw, can i add permissions parameter to restrict chat member? And show warning if it is not set, so i won't have to add params like can_send_audio there:
image

Badiboy:
> Btw, can i add permissions parameter to restrict chat member? And show warning if it is not set, so i won't have to add params like can_send_audio there

I propose to have full copy of parameters, it looks more consistent, but I don't insist, it's up to you.

coder2020official:
Cuz by copying those parameters we'll end up with more work afterwards... it'll be easier to break backward compatibility after some time ig

coder2020official:
If users will need new features from latest bot api update - they'll have to use method correctly :)

coder2020official:
what do u think?

telebot/__init__.py
```diff
@@ -2972,15 +2977,31 @@ def restrict_chat_member(
...
+                can_send_media_messages=can_send_media_messages,
+                can_send_polls=can_send_polls,
+                can_send_other_messages=can_send_other_messages,
+                can_add_web_page_previews=can_add_web_page_previews,
+                can_change_info=can_change_info,
+                can_invite_users=can_invite_users,
+                can_pin_messages=can_pin_messages
+            )
+            logger.warning(
+                'Please pass `telebot.types.ChatPermissions` object to `permissions` parameter instead of '
```
Badiboy, line 3000: I propose to use "Individual parameters are deprecated and will be removed, use 'permissions'" instead.

telebot/types.py
```diff
@@ -3156,6 +3162,14 @@ def __init__(self, can_send_messages=None, can_send_audios=None,
         self.can_send_video_notes: bool = can_send_video_notes
         self.can_send_voice_notes: bool = can_send_voice_notes
 
+        if can_send_media_messages is not None:
```
Badiboy, line 3165: Also "deprecated" warning here I think.

telebot/async_telebot.py
```diff
@@ -3784,6 +3785,10 @@ async def restrict_chat_member(
 
         Telegram documentation: https://core.telegram.org/bots/api#restrictchatmember
 
+        .. warning::
+            Please pass `telebot.types.ChatPermissions` object to `permissions` parameter instead of
```
Badiboy, line 3789: Same here.

Badiboy:
Just some textual refinement proposal.

coder2020official:
Done ig

Badiboy:
That's it .Approove. )

Badiboy (approved):

Badiboy:
@coder2020official Pin me when you decide that it's ready for release. Even if right now.

coder2020official:
Yeah i think we can release it now. But in changelog, we have to include a paragraph about deprecated things, smth like:

Few notices about deprecated things:
We recommend switching to proper functions, classes , parameters if you are using deprecated ones. Deprecated functions will be deleted soon.

About update backward compatibility
If you are using restrict_chat_member method, please pass permissions parameter instead of passing all boolean values for different parameters. Passing parameters one-by-one to function is now deprecated and will be removed soon


I think after releasing library i can go through already deprecated params,funcs and classes, and delete them. There are very old deprecated things in the lib which can be removed. Opinions @Badiboy ?

Badiboy:
> But in changelog, we have to include a paragraph about deprecated things

We can. But nobody reads release notes... 😔

> go through already deprecated params,funcs and classes, and delete them

Formally you may be right. But (see above) - nobody read notes and will have suprize... As far as all those deprecated things makes no harm to us and project - I propose to keep them as long as possible.

coder2020official:
Don't worry, I have already announced those things in chats. But, since nobody reads release note, i think we can keep those deprecated things for 2-3 releases, and add those notes about deprecated things in each release note. Then we can remove them :)

coder2020official:
We won't remove recently deprecated things, but, we will remove old deprecated things. There are loads of them.

PR #1917, opened 2023-02-15
Add a function to extract contents of entities from messages

S1RANN:
## Description
Add `extract_entity` in  `util.py` to extract contents of entities from messages.

## Describe your tests
Send a message and compare the extracted contents of returned entities to the correct ones

Python version:3.10

OS: Windows

## Checklist:

### DIFF
--- a/telebot/util.py
+++ b/telebot/util.py
@@ -274,6 +274,39 @@ def extract_arguments(text: str) -> str or None:
     result = regexp.match(text)
     return result.group(2) if is_command(text) else None
 
+def extract_entity(text: str, e: types.MessageEntity) -> str:
+    """
+    Returns the content of the entity.
+    
+    :param text: The text of the message the entity belongs to
+    :type text: :obj:`str`
+    
+    :param e: The entity to extract
+    :type e: :obj:`MessageEntity`
+    
+    :return: The content of the entity
+    :rtype: :obj:`str`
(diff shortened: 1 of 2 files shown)


Badiboy:
As for me, I do not understand the purpose of the function.
Can you provide some use cases for it?

S1RANN:
> As for me, I do not understand the purpose of the function. Can you provide some use cases for it?

Some apps create a share link which consists of a url and a description you can send to telegram, you can then let the bot grab some stuff like images from that. With this function you can acquire that url.

S1RANN:
> But... Message have entities...
> 
> image
> 
> And entity has everything parsed...
> 
> image
> 
> https://core.telegram.org/bots/api#messageentity

There are two types of url, one is `url` : https://github.com and the other one is `text_link`: [Github](https://github.com).
That url field is just for the second type.
For the first one you have to parse it yourself.

Badiboy:
Hmm.. That's true.

So, your function extratcs entity text by Message text and MessageEntity, right?

S1RANN:
> Hmm.. That's true.
> 
> So, your function extratcs entity text by Message text and MessageEntity, right?

Yes

Badiboy:
@coder2020official ?

coder2020official:
Looks fine

PR #1920, opened 2023-02-16
Fixed punctuation, added a new public bot

CommanderCRM:


Badiboy:
This place is not the collection of every single-finction-bot of everyones life.

PR #1921, opened 2023-02-17
Fix .webm upload

zeldpol:
No need to pass file content as a header, it causes the http error "431 Request Header Fields Too Large".


Badiboy:
Sounds reasonable. Looks like a copy-paste on update when new field was added to Bot API.

@coder2020official I confirm that it's right.

@zeldpol Could you please do the same update to [asyncio_helper.py](https://github.com/eternnoir/pyTelegramBotAPI/blob/master/telebot/asyncio_helper.py)?

coder2020official:
Agree

PR #1947, opened 2023-03-18
redesigned the antiflood method for guaranteed message delivery

alex75311:
## Description
redesigned the antiflood method for guaranteed message delivery

## Checklist:

### DIFF
--- a/telebot/util.py
+++ b/telebot/util.py
@@ -588,7 +588,7 @@ def webhook_google_functions(bot, request):
         return 'Bot ON'
 
 
-def antiflood(function: Callable, *args, **kwargs):
+def antiflood(function: Callable, *args, number_retries=5, **kwargs):
     """
     Use this function inside loops in order to avoid getting TooManyRequests error.
     Example:
@@ -602,6 +602,9 @@ def antiflood(function: Callable, *args, **kwargs):
     :param function: The function to call
     :type function: :obj:`Callable`
 
+    :param number_retries: Number of retries to send
+    :type function: :obj:int
+
     :param args: The arguments to pass to the function
     :type args: :obj:`tuple`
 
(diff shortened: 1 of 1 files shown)


coder2020official:
Hi!
Thank you for PR. We'll review it after bot api 6.6 update

alex75311:
the library uses multithreading. another thread manages to squeeze in and send, and part of the threads returns error 429 with a new timeout. my solution guarantees that the message will be sent\modified.

Badiboy (approved):

coder2020official:
Should be tested first

coder2020official:
the original function does not raise an error when the message is not delivered, i don't see any reason for redesigning cuz the previous version already handles 429 error which is the only error to be handled

alex75311:
above I described a case in which the current solution did not work for me, so I changed it to a solution that delivers 100% messages.

telebot/util.py
```diff
@@ -613,14 +613,14 @@ def antiflood(function: Callable, *args, **kwargs):
...
     from time import sleep
 
-    try:
-        return function(*args, **kwargs)
-    except ApiTelegramException as ex:
-        if ex.error_code == 429:
-            sleep(ex.result_json['parameters']['retry_after'])
+    while True:
+        try:
             return function(*args, **kwargs)
```
alex75311, line 618: if error 429 is returned here again, then it is not processed in any way

coder2020official:
But raising the error is a breaking change

Badiboy:
> But raising the error is a breaking change

It raised error before. With this update it will raise it more rare.

From my side I cannot accept infinite loops in processing. Various things may happen, thus infinite loop will bring to stall. Loop may exist but it should be finite. May be with configurable (by parameter) number of retries.
@alex75311 👆

alex75311:
I added 5 attempts, that should be enough

Badiboy:
@coder2020official Looks fine for me now.

Badiboy:
@coder2020official Reasonable.

@alex75311 Please, set in line 628 "raise" instead of "return ...".

coder2020official:
i am talking about last line:
image
why is an error being raised? is it intended?

alex75311:
if we are talking about line 6 of your screenshot, then this exception is called if the function returned an error other than 429

coder2020official:
Yes, is it intended? Before it never raised an error

alex75311:
look carefully at how it was before

Badiboy:
> Yes, is it intended? Before it never raised an error

image

Absolutely the same.

coder2020official:
Fine then, I probably skipped the last part

coder2020official (approved):

PR #1951, opened 2023-03-23
Fix documentation for `InlineKeyboardMarkup` and `quick_markup`

mikelei8291:
## Description

- The documentation for `InlineKeyboardMarkup` was outdated so I updated it to be in sync with what `quick_markup` provides.
- Removed the extra `: ` characters at the end of `telebot/util.py`:
  https://github.com/eternnoir/pyTelegramBotAPI/blob/da57174635cf2b22281f3d9f024fb540d0b9246e/telebot/util.py#L427
- Fixed the parameters documentation for `InlineKeyboardMarkup` and rewrite the description of the `row_width` parameter for both the class itself and the function `quick_markup`.

## Describe your tests

I built the docs using Sphinx and checked the HTML output, and the output was correct.

Python version: 3.10.4
Sphinx version: 6.1.3
OS: Ubuntu 20.04.5 LTS

## Checklist:

### DIFF
--- a/telebot/types.py
+++ b/telebot/types.py
@@ -2432,23 +2432,29 @@ class InlineKeyboardMarkup(Dictionaryable, JsonSerializable, JsonDeserializable)
     This object represents an inline keyboard that appears right next to the message it belongs to.
 
     .. note::
-        It is recommended to use :meth:`telebot.service_utils..quick_markup` instead.
+        It is recommended to use :meth:`telebot.util.quick_markup` instead.
 
     .. code-block:: python3
         :caption: Example of a custom keyboard with buttons.
 
-        from telebot.service_utils..import quick_markup
+        from telebot.util import quick_markup
 
-        markup = quick_markup(
-            {'text': 'Press me', 'callback_data': 'press'},
-            {'text': 'Press me too', 'callback_data': 'press_too'}
-        )
+        markup = quick_markup({
+            'Twitter': {'url': 'https://twitter.com'},
+            'Facebook': {'url': 'https://facebook.com'},
+            'Back': {'callback_data': 'whatever'}
+        }, row_width=2)
+        # returns an InlineKeyboardMarkup with two buttons in a row, one leading to Twitter, the other to facebook
+        # and a back button below
 
     Telegram Documentation: https://core.telegram.org/bots/api#inlinekeyboardmarkup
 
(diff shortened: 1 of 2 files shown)


Badiboy (approved):

telebot/types.py
```diff
@@ -2432,23 +2432,29 @@ class InlineKeyboardMarkup(Dictionaryable, JsonSerializable, JsonDeserializable)
...
+            'Facebook': {'url': 'https://facebook.com'},
+            'Back': {'callback_data': 'whatever'}
+        }, row_width=2)
+        # returns an InlineKeyboardMarkup with two buttons in a row, one leading to Twitter, the other to facebook
+        # and a back button below
 
     Telegram Documentation: https://core.telegram.org/bots/api#inlinekeyboardmarkup
 
-    :param inline_keyboard: :obj:`list` of button rows, each represented by an :obj:`list` of 
+    :param keyboard: :obj:`list` of button rows, each represented by an :obj:`list` of
```
coder2020official, line 2452: @Badiboy [doc](https://core.telegram.org/bots/api#inlinekeyboardmarkup) says that it should be inline_keyboard. Ig we should have another PR changing keyboard param to inline_keyboard

coder2020official (changes requested):

telebot/types.py
mikelei8291: Do I need to change this back to `inline_keyboard`? However, I do notice other inconsistencies of parameter names between the library and the Telegram API documentation. For example, the `InlineQueryResultPhoto` class has parameter `thumb_url`, while the name of the field in the API documentation is `thumbnail_url` instead.

Badiboy: > @Badiboy [doc](https://core.telegram.org/bots/api#inlinekeyboardmarkup) says that it should be inline_keyboard.

1. He just align the **description** with **actual name**.
image
And this is correct update.

2. doc have nothing about **class constructor parameters**. It describes **fields**. So construtor may have parameter "keyboard" and it is more laconic then "inline_keyboard". And "inline_" prefix is absolete because you create InlineKeyboard: it's absolutely clear that keyboard is inline.

3. Anyway it's not possible to change keyboard to inline_keyboard due to compatibility. And I see no vvaluable reason to create new parameter and deprecate old.

coder2020official:
Looks good

PR #1964, opened 2023-04-14
Add state_list to StatesGroup

Cub11k:
## Description

## Describe your tests
How did you test your change?

Python version:

OS:

## Checklist:


coder2020official:
Why not to add it during the initialisation of the subclass and then return the list?

Cub11k:
Definitely better, agreed

coder2020official (approved):

Badiboy (approved):

PR #1992, opened 2023-05-25
Update `apihelper.py`

TECH-SAVVY-GUY:
Add support for test mode in the Bot API

image

https://core.telegram.org/bots/webapps#testing-web-apps

### DIFF
--- a/telebot/apihelper.py
+++ b/telebot/apihelper.py
@@ -45,7 +45,7 @@ CUSTOM_SERIALIZER = None
 CUSTOM_REQUEST_SENDER = None
 
 ENABLE_MIDDLEWARE = False
-
+TEST_MODE = False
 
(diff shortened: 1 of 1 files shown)


coder2020official:
@Badiboy it is still possible to set test env by manually setting the URL, but I think PR is still fine. But it would be maybe better to rename it to TEST_API_URL(?)

Badiboy:
Do not see any value in this PR.

Setting
`API_URL = "https://api.telegram.org/bot{0}/test/{1}"`

Solve the task.

TECH-SAVVY-GUY:
Added this because an entire section was dedicated on testing web apps in the Telegram API.

Badiboy:
Do not see any value in this PR. Use API_URL.

PR #2008, opened 2023-07-08
Fix issue with post_process in async not receiving the error

coder2020official:
## Description

## Describe your tests
How did you test your change?

Python version:

OS:

## Checklist:


coder2020official:
Sync is fine:
https://github.com/eternnoir/pyTelegramBotAPI/blob/master/telebot/__init__.py#L6829

PR #2009, opened 2023-07-08
Fixes #1944: uploading file from memory

### LINKED ISSUE #1944: IOBase does not contain a name attribute
https://github.com/eternnoir/pyTelegramBotAPI/blob/5d9a76b0dd0d3ee88c5c9d1329c06b24fdc4457b/telebot/types.py#L7056

Unable to load file from memory, e.g. from io.BytesIO

coder2020official:
@Badiboy i'm not really sure about this: Basically i am creating random string for file name in InputFile for memory-uploaded files. Do you think it is fine?

### DIFF
--- a/telebot/types.py
+++ b/telebot/types.py
@@ -7250,7 +7250,7 @@ class InputFile:
             _file = open(file, 'rb')
             return _file, os.path.basename(_file.name)
         elif isinstance(file, IOBase):
-            return file, os.path.basename(file.name)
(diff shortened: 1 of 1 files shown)


Badiboy:
It's a bit strange (I think that it's more convenient to use some default name) but should work...

coder2020official:
I guess using a default name is wrong... so do we merge?

Badiboy:
do merge )

PR #2017, opened 2023-07-14
Fix backslash (\) issue in escape_markdown

Artin-GH:
## Fix backslash (\) issue in escape_markdown
When I was using backslash in `telebot.formatting.escape_markdown`, I was getting an error from the Telegram API.
The regex pattern inside the `escape_markdown` function was missing a `backslash` within the brackets.

### DIFF
--- a/telebot/formatting.py
+++ b/telebot/formatting.py
@@ -61,8 +61,8 @@ def escape_markdown(content: str) -> str:
     :rtype: :obj:`str`
     """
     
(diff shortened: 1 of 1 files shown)


Badiboy:
Could you please provide usecase(s) where you got the error that will be fixed with your patch?

Badiboy:
Thank you, looks reasonable.

@coder2020official It's your function - please, confirm update.

coder2020official:
Sorry for a late response

PR #2024, opened 2023-07-24
added posibility to handle exception on RequestTimeout

TheWCKD:
## Description
I wanted to increment an error counter whenever the RequestTimeout occurs as an exception via the exception_handler class

## Describe your tests
How did you test your change?

Tested on my own project in which I frequently use telebot

Python version: 3.11

OS: Linux Mint

## Checklist:


Badiboy:
@coder2020official Moved from sync version.

coder2020official:
Will get back soon

coder2020official:
Thanks

PR #2026, opened 2023-07-27
add-new-example

Vinnybrunn00:
## Description
Include changes, new features and etc: I made the change in the [examples](https://github.com/eternnoir/pyTelegramBotAPI/tree/master/examples) folder

## Describe your tests
- I created an example that captures all new member events for all groups. I looked for something similar in the example folder but I didn't find it, so I decided to make an example and contribute, this can be useful for those starting to learn the API

Python version: 3.11

OS: **Linux**

## Checklist:

### DIFF
--- a/examples/new_chat_members_example.py
+++ b/examples/new_chat_members_example.py
@@ -1,0 +1,17 @@
+from telebot.async_telebot import AsyncTeleBot
+from asyncio import run
+
+bot = AsyncTeleBot('YOUR TOKEN')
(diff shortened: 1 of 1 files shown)


coder2020official:
Usage of message.json is not appropriate for an example

Badiboy:
1. Async examples are placed in pyTelegramBotAPI/examples/asynchronous_telebot
2. Working with message as json is extremely auful and is a VERY BAD example. Before creating examples you should first take more on best practices...

PR #839, opened 2020-05-12
add PollAnswer, poll_answer_handler

drforse:
+ add PollAnswer, poll_answer_handler
+ make User Serializable and Dictionaryble
+ some pep fixes

PR #2030, opened 2023-08-18
Bot API 6.8: Stories, voting as a chat, and other minor changes

coder2020official:
### **Bot API 6.8**

- [x] Added the field story to the class [Message](https://core.telegram.org/bots/api#message) for messages with forwarded stories. Currently, it holds no information.

- [x] Added the field voter_chat to the class [PollAnswer](https://core.telegram.org/bots/api#pollanswer), to contain channel chat voters in [Polls](https://core.telegram.org/bots/api#poll). For backward compatibility, the field user in such objects will contain the user 136817688 ([@Channel_Bot](https://t.me/Channel_Bot)).
- [x] Added the field emoji_status_expiration_date to the class [Chat](https://core.telegram.org/bots/api#chat).
- [x] Added the method [unpinAllGeneralForumTopicMessages](https://core.telegram.org/bots/api#unpinallgeneralforumtopicmessages).
- [x] Increased to 512 characters the maximum length of the startapp parameter in direct Web App **links(NOT NEEDED)**

### DIFF
--- a/telebot/types.py
+++ b/telebot/types.py
@@ -821,6 +821,9 @@ class Message(JsonDeserializable):
     :param sticker: Optional. Message is a sticker, information about the sticker
     :type sticker: :class:`telebot.types.Sticker`
 
+    :param story: Optional. Message is a forwarded story
+    :type story: :class:`telebot.types.Story`
+
     :param video: Optional. Message is a video, information about the video
     :type video: :class:`telebot.types.Video`
 
@@ -1177,6 +1180,9 @@ class Message(JsonDeserializable):
         if 'chat_shared' in obj:
             opts['chat_shared'] = ChatShared.de_json(obj['chat_shared'])
             content_type = 'chat_shared'
+        if 'story' in obj:
+            opts['story'] = Story.de_json(obj['story'])
+            content_type = 'story'
         return cls(message_id, from_user, date, chat, content_type, opts, json_string)
 
     @classmethod
@@ -1274,9 +1280,18 @@ class Message(JsonDeserializable):
         self.write_access_allowed: Optional[WriteAccessAllowed] = None
         self.user_shared: Optional[UserShared] = None
         self.chat_shared: Optional[ChatShared] = None
+        self.story: Optional[Story] = None
         for key in options:
             setattr(self, key, options[key])
         self.json = json_string
+
+    @property
+    def is_forwarded_story(self):
+        """
+        Returns True if the message is a forwarded story.
+        """
+        # TODO: In future updates this propery might require changes.
+        return True if self.story is not None else False
 
     def __html_text(self, text, entities):
         """
@@ -7743,3 +7758,18 @@ class InlineQueryResultsButton(JsonSerializable, Dictionaryable):
     
     def to_json(self) -> str:
         return json.dumps(self.to_dict())
+
+
+class Story(JsonDeserializable):
+    """
+    This object represents a message about a forwarded story in the chat.
+    Currently holds no information.
+    """
+
+    @classmethod
+    def de_json(cls, json_string):
+        return cls()
+
+    def __init__(self) -> None:
+        pass
+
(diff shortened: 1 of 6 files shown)


coder2020official:
@badiboy can you clarify one thing: what is the purpose of to_dict in types.PollAnswer? Check previous commit where I added voter_chat, and if to_dict is necessary I have to add to_dict for types.Chat. I have checked api doc and I see no need for to_dict for PollAnswer, since it is an update. Or did I miss something?

coder2020official:
I guess I am done except for the PollAnswer issue I described above.

Badiboy:
> @Badiboy can you clarify one thing: what is the purpose of to_dict in types.PollAnswer?

It was long long time ago. :)
https://github.com/eternnoir/pyTelegramBotAPI/pull/839

They are not needed in library. And I suppose they are not used by anybody.
I prefer them to stay as far as they are already exists, but if you **hardly** want to remove it - I think they can be removed.

Badiboy:
The same you can check user.to_dict/to_json if you want to, I do not remember what's with these methods usage.

telebot/types.py
```diff
@@ -7743,3 +7758,18 @@ def to_dict(self) -> dict:
...
+
+class Story(JsonDeserializable):
+    """
+    This object represents a message about a forwarded story in the chat.
+    Currently holds no information.
+    """
+
+    @classmethod
+    def de_json(cls, json_string):
+        return cls()
```
Badiboy, line 7783: image

Generally we proceed json in this way. Currently it does not matter, but I suppose it's better to have standart code signature even here.

telebot/types.py
```diff
@@ -1274,10 +1280,19 @@ def __init__(self, message_id, from_user, date, chat, content_type, options, jso
         self.write_access_allowed: Optional[WriteAccessAllowed] = None
         self.user_shared: Optional[UserShared] = None
         self.chat_shared: Optional[ChatShared] = None
+        self.story: Optional[Story] = None
         for key in options:
             setattr(self, key, options[key])
         self.json = json_string
 
+    @property
+    def is_forwarded_story(self):
```
Badiboy, line 1294: See no reason in appearance of this property... No is_anumation, is_video, is_xxx - why now is_story?

telebot/types.py
```diff
@@ -6710,12 +6713,16 @@ def de_json(cls, json_string):
         if (json_string is None): return None
         obj = cls.check_json(json_string)
         obj['user'] = User.de_json(obj['user'])
+        if 'voter_chat' in obj:
```
Badiboy, line 6721: User is now optional and should be processed the same as voter_chat with if

telebot/types.py
```diff
@@ -6710,12 +6713,16 @@ def de_json(cls, json_string):
...
+            obj['voter_chat'] = Chat.de_json(obj['voter_chat'])
         return cls(**obj)
 
-    def __init__(self, poll_id, user, option_ids, **kwargs):
+    def __init__(self, poll_id, user, option_ids, voter_chat=None, **kwargs):
```
Badiboy, line 6725: And user is optional here also... Seems we need to change init signature.

telebot/types.py
```diff
@@ -6710,12 +6713,16 @@ def de_json(cls, json_string):
...
         self.poll_id: str = poll_id
         self.user: User = user
```
Badiboy, line 6727: And here...

Badiboy:
@coder2020official I submitted review. If you don't see it let me know, means I did something wrong again )

coder2020official:
@badiboy i guess I am done, I removed the is_forwarded_story option(found it useless after ur comment), made user optional, improved new story class to check for json. Check the previous commit for more

coder2020official:
Tip: You can also suggest a commit and I will be able to commit it right away
image

Badiboy (approved):

PR #2041, opened 2023-09-08
Add redis_url to StateRedisStorage to allow initialization by URL

Cub11k:
## Description
Useful for rediss:// and unix:// schemes

## Describe your tests
How did you test your change?

Python version:

OS:

## Checklist:
- [ ] I added/edited example on new feature/change (if exists)
- [x] My changes won't break backward compatibility
- [x] I made changes both for sync and async

PR #2042, opened 2023-09-08
The ability to work with from_url() method for Redis

Vokler:
## Description
Added the ability to work with Redis URL. Now we can pass URL of Redis and from_url() method will be used for creating ConnectionPool class

## Describe your tests
How did you test your change?
Test on my bot with URL and without URL

Python version: 3.10

OS: MacOS 13.4.1 (c) M1

## Checklist:


Badiboy:
https://github.com/eternnoir/pyTelegramBotAPI/pull/2041

One is enough. :)

PR #2062, opened 2023-10-18
Update README.md

alexsuhor:
## Description
should have to functions -> should have two functions

## Describe your tests
How did you test your change?

Python version:

OS:

## Checklist:


Badiboy:
Thank you.

PR #2078, opened 2023-11-12
Update InlineKeyboardMarkup in types.py

dembelson:
Adding special method __add__ to class InlineKeyboardMarkup. To better manage markups when using quick_markup. Without resorting to further use of InlineKeyboardButton to add elements.

### DIFF
--- a/telebot/types.py
+++ b/telebot/types.py
@@ -2583,6 +2583,27 @@ class InlineKeyboardMarkup(Dictionaryable, JsonSerializable, JsonDeserializable)
         json_dict['inline_keyboard'] = [[button.to_dict() for button in row] for row in self.keyboard]
         return json_dict
 
+    def __add__(self, other: InlineKeyboardMarkup) -> InlineKeyboardMarkup:
+        """
+        Using: final_InlineKeyboardMarkup = InlineKeyboardMarkup_obj(rsize=5) + InlineKeyboardMarkup_obj(rsize=3)
+
+        This improvement using to better manage buttons using quick_markup.
+        So using this you could make first initial quick markup. And the add anothers to it not using self.add(InlineKeyBoardButton).
+        
+        Adding one markup to another with different row_sizes:
+        If first mark row_width smaller than in the second then second elements will stand as in first mark.
+
+        :param other: InlineKeyboardMarkup
+        :return: InlineKeyboardMarkup
+        """
+        if not isinstance(other, InlineKeyboardMarkup):
+            raise TypeError('Adding argument is not InlineKeyboardMarkup object!')
+        other_mark_dict = other.to_dict()['inline_keyboard']
+        btns = [[InlineKeyboardButton(**x) for x in other_mark_btns] for other_mark_btns in other_mark_dict]
+        for i in btns:
+            self.add(*i)
+        return self
+
 
 class InlineKeyboardButton(Dictionaryable, JsonSerializable, JsonDeserializable):
     """
--- a/telebot/types.py
+++ b/telebot/types.py
@@ -2583,16 +2583,11 @@ class InlineKeyboardMarkup(Dictionaryable, JsonSerializable, JsonDeserializable)
         json_dict['inline_keyboard'] = [[button.to_dict() for button in row] for row in self.keyboard]
         return json_dict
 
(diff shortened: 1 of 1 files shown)


Badiboy:
1. I see the comment in function but do not understand it. Please clearly provide **here** the reason, how the given example will work (what result will give) and the real use cases where it may be required.

2. For me personally overridint system method __add__ will be confising for such class like this one. I prefer to extend "add" method to handle adding of ready markup. Your case will be:
final_InlineKeyboardMarkup = InlineKeyboardMarkup_obj(rsize=5)
final_InlineKeyboardMarkup.add(InlineKeyboardMarkup_obj(rsize=3))

dembelson:
> 1. I see the comment in function but do not understand it. Please clearly provide **here** the reason, how the given example will work (what result will give) and the real use cases where it may be required.
> 2. For me personally overridint system method **add** will be confising for such class like this one. I prefer to extend "add" method to handle adding of ready markup. Your case will be:
>    final_InlineKeyboardMarkup = InlineKeyboardMarkup_obj(rsize=5)
>    final_InlineKeyboardMarkup.add(InlineKeyboardMarkup_obj(rsize=3))

1.Example:
```python
qmark= {author: {'callback_data': author} for author in authors_list}
x = quick_markup(qmark, row_width=4)
menu_qmark = {'<<': {'callback_data': 'page_back'}, 'Cancel': {'callback_data': 'cancel'}}, 1, '>>': {'callback_data': 'page_forward'}}
y = quick_markup(menu_qmark, row_width=3)
final_qmark = x+y
await bot.send_message(..reply_markup=final_mark)
``` 
Message marks view will be like:
"|author1 | author2| author3| author4|
 |        author5     |       author6           | 
 |     <<     |       Cancel       |      >>     |"


2.I've been watch to calendar example in ".examples/asynchronous_telebot/callback_data_examples/advanced_calendar_example/" and there in "keyboards.generate_calendar_days" using self.add(InlineKeyboardButton(..)) for buttons formatting. But InlineKeyboardMarkup pushes to use telebot.util.quick_markup (I suppose it pushing to make using InlineKeyboardMarkup more easier). And quick_markup let's only once to set row size and then you'll need to self.add InlineKeyboardButton separately again when you want to add and format new buttons.

I agree that it is not necessary to use addition(+) with using a special method, but simply to write a separate class method. 
Or it might be better to add isinstance checking to "add" method. 
(InlineKeyboardButton or another InlineKeyboardMarkup when adding to existing InlineKeyboardMarkup).

Badiboy:
> "|author1 | author2| author3| author4|
> | author5 | author6 |
> | << | Cancel | >> |"

Got it. It's fine. I propose to simply add new method like "add_markup" (I prefer this) or may be "extend".

Badiboy:
Looks fine for me. Let's wait some time for @coder2020official , may be he will have time to take a look.

dembelson:
> Looks fine for me. Let's wait some time for @coder2020official , may be he will have time to take a look.

But when first row row_size (when you first use quick_markup) is 1. And you need next to be 3. They all will be 1.
|author1|
|author2|
|<<|
|Cancel|
|>>|

To get around this I use:
mark = quick_markup(...row_width=1)
mark.row_width = 3
menu_mark = quick_markup(...row_width=3)
final = mark.add_markup(menu_mark)
and it's work fine.
|author1|
|author2|
|<<|Cancel|>>|

Perhaps needs to add it to the method:
if self.row_width < other.row_width:
    self.row_width = other.row_width

Badiboy:
You cannot change row_width of existsing markup. It's unpredictable behaviour.

Badiboy:
Use "row" instead of "add"

dembelson:
> Use "row" instead of "add"

You're abolutely right!

Also I think that it will much better to use
```python
for btns in other.keyboard:
   self.row(*btns)
```  

instead of 

```python
other_mark_dict = other.to_dict()['inline_keyboard']
btns = [[InlineKeyboardButton(**x) for x in other_mark_btns] for other_mark_btns in other_mark_dict]
for i in btns:
    self.row(*i)
```

Badiboy:
> for btns in other.keyboard:
>     self.keyboard.append(btns)

Now you can see that this method is not needed in fact. The PR can be closed.

PR #2083, opened 2023-11-20
Fix telebot.types.Message.html_text may contain unescaped HTML characters.

EgorKhabarov:
## Description
A Telegram user could send HTML characters without markdown markup and they could end up in html_text.

## Describe your tests
The user sends `<b>b</b> <i>i</i>` (without markdown characters) and as an html string I expect to receive the escaped characters `&lt;b&gt;b&lt;/b&gt; &lt;i&gt;i&lt;/i&gt;`
If the user sends `**b** <b>b</b> <i>i</i>` (with markdown characters) then I get `<b>b</b> &lt;b&gt;b&lt;/ b&gt; &lt;i&gt;i&lt;/i&gt;`.

Python version: 3.11.6

OS: Windows 10

## Checklist:

### DIFF
--- a/telebot/types.py
+++ b/telebot/types.py
@@ -1294,7 +1294,7 @@ class Message(JsonDeserializable):
     def __html_text(self, text, entities):
         """
         Author: @sviat9440
-        Updaters: @badiboy
+        Updaters: @badiboy, @EgorKhabarov
         Message: "*Test* parse _formatting_, [url](https://example.com), [text_mention](tg://user?id=123456) and mention @username"
 
         .. code-block:: python3
@@ -1314,7 +1314,7 @@ class Message(JsonDeserializable):
         """
 
         if not entities:
-            return text
+            return text.replace("&", "&amp;").replace("<", "&lt;").replace(">", "&gt;")
 
         _subs = {
             "bold": "<b>{text}</b>",
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -295,7 +295,15 @@ def test_message_entity():
     message_4 = types.Update.de_json(sample_string_4).message
     assert message_4.html_text == '<s><u><i><b>aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa</b></i></u></s><tg-emoji emoji-id="5456188142006575553">😋</tg-emoji><tg-emoji emoji-id="5456188142006575553">😋</tg-emoji>'
 
-    
+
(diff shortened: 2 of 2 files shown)


Badiboy:
> The user sends <b>b</b> <i>i</i> (without markdown characters) and as an html string I expect to receive the escaped characters &lt;b&gt;b&lt;/b&gt; &lt;i&gt;i&lt;/i&gt;

I absolutely do not expect the same. User send - you receive. If you want to - you can replace them on sending. In **your** bot. You cannot modify the user input just because **you** like it.

EgorKhabarov:
If there is no Markdown markup in the message, `message.html_text` may contain user-inputted `<` or `>` characters
that are not processed by pyTelegramBotAPI. In the bot code, an additional check for the presence of `message.entities` will be necessary. I think it's logical that only valid HTML characters will end up in `message.html_text`.

Badiboy:
For at least 5 years for everyone it was logical enough "as is". It will stay the same logical further.

EgorKhabarov:
photo_2023-11-29_11-20-41

```
from telebot import TeleBot
from telebot.types import Message
from telebot.apihelper import ApiTelegramException

bot = TeleBot(token)
bot.parse_mode = "html"

@bot.message_handler()
def handler(message: Message):
    try:
        bot.send_message(message.chat.id, message.html_text)
    except ApiTelegramException as e:
        bot.send_message(message.chat.id, str(e))

bot.infinity_polling()
```

@Badiboy, Is this logical behavior?

You need to write extra code for it to work correctly.

photo_2023-11-29_11-29-03

```
import html

from telebot import TeleBot
from telebot.types import Message
from telebot.apihelper import ApiTelegramException

bot = TeleBot(token)
bot.parse_mode = "html"

@bot.message_handler()
def handler(message: Message):
    html_text = message.html_text if message.entities else html.escape(message.html_text)
    try:
        bot.send_message(message.chat.id, html_text)
    except ApiTelegramException as e:
        bot.send_message(message.chat.id, str(e))

bot.infinity_polling()
```

Badiboy:
If you use parse mode "HTML" - yes, it's logical behaviour that all HTML-like tags are assumed to be parsed. If you do not need  HTML symbols to be parsed - escape them.

EgorKhabarov:
I need to accept text with markup from the user, and then display it by turning HTML into markdown.
photo_2023-11-29_11-53-37

Here https://github.com/eternnoir/pyTelegramBotAPI/blob/master/telebot/types.py#L1345 HTML characters are escaped.
Why aren't they escaped without message.entities?
I don't understand why this seems logical to you.
I'm forced to write additional code that looks like a crutch.
Instead of calling one function to convert html to markdown I should check message.entities and call html.escape...

EgorKhabarov:
I really don't understand why these two messages would look the same in message.html_text without any additional processing.
image

Badiboy:
I already gave you all answers. The fact you don't like them does not mean I should give more.

PR #2095, opened 2023-12-09
Specified correct return type

AsMaNick:
## Description
Specified the correct return type of the `copy_message` method.

## Describe your tests
Only documentation was changed, no tests were needed.

## Checklist: 


Badiboy:
Thank you!

Issue #1853, opened 2022-12-29
Usage of black code formatter


coder2020official:
I'd recommend using [black code formatter](https://github.com/psf/black) in the project.

**What problem does it solve?**
It solves pep 8 issues in code. Since library has a lot of pep 8 issues, I think it'd be better to integrate it into our project.

**How does it work?**
I can install it, apply changes to our project, then make a PR. 

**Background**
I've already recommended using it, but by that time it wasn't stable and was in beta testing. However, currently, it is stable and it is used by many projects:
<img width="866" alt="image" src="https://user-images.githubusercontent.com/63740404/210004124-0322b932-8732-4397-be22-7f4caecb5ea4.png">

Thoughts, @Badiboy ?

coder2020official:
For more:
https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html

coder2020official:
We have tests in the repo, so if there're some issues in syntax, tests are gonna fail. Nothing to worry about.

Badiboy:
> Thoughts, @Badiboy ?

My thoughts did not changed since the last time you propose it. I don't see it brings something valuable, so it will be just the waste of time.

However if you want to have fun bringing things that are valueless to my oppininon - do it. You are the active supporter of the library, you have enough rights to do it.

coder2020official:
It is not valueless. We have to keep project clean. By the way, it is not me who cleans whole code myself - it is an automated system, so it is not waste of time.

PR #2111, opened 2024-01-03
Post-huge update fix #2: deprecations

### LINKED ISSUE #2: Can't kill the echo bot example
First, great work! It has been very useful and fun to me.
I copy/pasted the example, but can't be interrupted with the KeyboardInterrupt signal on Gentoo.
IMHO one possible fix could be change the main loop at the end:
try:
    while 1:
        time.sleep(20)
except KeyboardInterrupt:
    tb.stop_polling()

Badiboy:
## Description
Deprecation fixes

### DIFF
--- a/telebot/types.py
+++ b/telebot/types.py
@@ -785,285 +785,285 @@ class WebAppData(JsonDeserializable, Dictionaryable):
 # noinspection PyUnresolvedReferences
 class Message(JsonDeserializable):
     """
-        This object represents a message.
+    This object represents a message.
 
-        Telegram Documentation: https://core.telegram.org/bots/api#message
+    Telegram Documentation: https://core.telegram.org/bots/api#message
 
-        :param message_id: Unique message identifier inside this chat
-        :type message_id: :obj:`int`
+    :param message_id: Unique message identifier inside this chat
+    :type message_id: :obj:`int`
 
-        :param message_thread_id: Optional. Unique identifier of a message thread to which the message belongs; for supergroups only
-        :type message_thread_id: :obj:`int`
+    :param message_thread_id: Optional. Unique identifier of a message thread to which the message belongs; for supergroups only
+    :type message_thread_id: :obj:`int`
 
-        :param from_user: Optional. Sender of the message; empty for messages sent to channels. For backward compatibility, the
-            field contains a fake sender user in non-channel chats, if the message was sent on behalf of a chat.
-        :type from_user: :class:`telebot.types.User`
+    :param from_user: Optional. Sender of the message; empty for messages sent to channels. For backward compatibility, the
+        field contains a fake sender user in non-channel chats, if the message was sent on behalf of a chat.
+    :type from_user: :class:`telebot.types.User`
 
-        :param sender_chat: Optional. Sender of the message, sent on behalf of a chat. For example, the channel itself for
-            channel posts, the supergroup itself for messages from anonymous group administrators, the linked channel for
-            messages automatically forwarded to the discussion group. For backward compatibility, the field from contains a
-            fake sender user in non-channel chats, if the message was sent on behalf of a chat.
-        :type sender_chat: :class:`telebot.types.Chat`
+    :param sender_chat: Optional. Sender of the message, sent on behalf of a chat. For example, the channel itself for
+        channel posts, the supergroup itself for messages from anonymous group administrators, the linked channel for
+        messages automatically forwarded to the discussion group. For backward compatibility, the field from contains a
+        fake sender user in non-channel chats, if the message was sent on behalf of a chat.
+    :type sender_chat: :class:`telebot.types.Chat`
 
-        :param date: Date the message was sent in Unix time
-        :type date: :obj:`int`
+    :param date: Date the message was sent in Unix time
+    :type date: :obj:`int`
 
-        :param chat: Conversation the message belongs to
-        :type chat: :class:`telebot.types.Chat`
+    :param chat: Conversation the message belongs to
+    :type chat: :class:`telebot.types.Chat`
 
-        :param forward_from: deprecated.
-        :type forward_from: :class:`telebot.types.User`
+    :param forward_from: deprecated.
+    :type forward_from: :class:`telebot.types.User`
 
-        :param forward_from_chat: deprecated.
-        :type forward_from_chat: :class:`telebot.types.Chat`
+    :param forward_from_chat: deprecated.
+    :type forward_from_chat: :class:`telebot.types.Chat`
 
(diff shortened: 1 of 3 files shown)


Badiboy:
(Not finished: more deprecation fixes are planned).

Badiboy:
@coder2020official I'm done. 
.

> Improved documentation for the field message of the class [callbackQuery](https://core.telegram.org/bots/api#callbackquery) and the field pinned_message of the class [Message](https://core.telegram.org/bots/api#message) by adding the classes [MaybeInaccessibleMessage](https://core.telegram.org/bots/api#maybeinaccessiblemessage) and [InaccessibleMessage](https://core.telegram.org/bots/api#inaccessiblemessage).

Is also done.

coder2020official:
image
Why extra spaces? If you really want a formatting style - use black

coder2020official:
I have previously suggested such approach
https://github.com/eternnoir/pyTelegramBotAPI/issues/1853

This PR seems to be mostly about style fixes. It would be better to leave style fixes for another PR.

Your implementation of inaccessible message makes sense, that part I approve

Badiboy:
> Why extra spaces

By mistake. It was occasionally formated, I did not paid attention.

Badiboy:
> This PR seems to be mostly about style fixes.

It is not. I MAY fix some style only as a "supporting" activity.

coder2020official:
Don't worry about style, spaces, etc - I will use black for the whole project if you approve

coder2020official:
Can you also please obey the order for each parameter in description? I get the order from the api doc, it is easier to navigate and stuff

coder2020official:
All of these were generated using api doc parsing. That is why most types have "ordered" parameters according to the doc

Badiboy:
> Don't worry about style, spaces, etc - I will use black for the whole project if you approve

I still don't think we need total redesign.

coder2020official:
I may update the doc strings in future when I find my old script or have time to recreate one

Badiboy:
> It's better if you take a look at your changes. I have seen some in message class, didn't check all classes though

I checked and didn't find **any** reordering made by me. Please, point exactly if I'm wrong.

coder2020official (commented):
This is all I could spot. And regarding the parameters order - sorry about that, I see nothing regarding order changed. 

Some questions:
- Did you make changes to async descriptions too? Would be nice if you could
- I see you are passing parameters by their name for init file. Did you check carefully so there are no typos?

That is it i guess, otherwise LGTM

coder2020official:
Oh wait, my comments got removed.

telebot/types.py
```diff
@@ -7703,7 +7788,7 @@ def de_json(cls, json_string):
         obj = cls.check_json(json_string)
         return cls(**obj)
 
-    def __init__(self, request_id: int, chat_id: int) -> None:
+    def __init__(self, request_id: int, chat_id: int, **kwargs) -> None:
```
coder2020official, line 7791: What is the purpose of kwargs when we don't use it?

telebot/types.py
```diff
@@ -957,50 +954,50 @@ class Message(JsonDeserializable):
...
+        identifier. This number may have more than 32 significant bits and some programming languages may have
+        difficulty/silent defects in interpreting it. But it has at most 52 significant bits, so a signed 64-bit integer or
         double-precision float type are safe for storing this identifier.
     :type migrate_from_chat_id: :obj:`int`
 
-    :param pinned_message: Optional. Specified message was pinned. Note that the Message object in this field will not 
+    :param pinned_message: Optional. Specified message was pinned. Note that the Message object in this field will not
         contain further reply_to_message fields even if it is itself a reply.
-    :type pinned_message: :class:`telebot.types.Message`
+    :type message: :class:`telebot.types.Message` or :class:`telebot.types.InaccessibleMessage`
```
coder2020official, line 985: Supposed to be pinned_message for type
```suggestion
    :type pinned_message: :class:`telebot.types.Message` or :class:`telebot.types.InaccessibleMessage`
```

telebot/types.py
```diff
@@ -8050,7 +8137,7 @@ class ReactionTypeCustomEmoji(ReactionType):
     :rtype: :class:`ReactionTypeCustomEmoji`
     """
 
-    def __init__(self, custom_emoji: str) -> None:
+    def __init__(self, custom_emoji: str, **kwargs) -> None:
```
coder2020official, line 8140: btw, custom_emoji should be turned to custom_emoji_id:
https://core.telegram.org/bots/api#reactiontypecustomemoji

coder2020official (changes requested):

Badiboy:
> * Did you make changes to async descriptions too? Would be nice if you could

For this update - no. There are no bugfixes, so these updates are not critical. If you like - you can bring them to Async or not, it's up to you as the Async version designer. Currently I'm not able to spend more days on Async: I overtake the available time to effort on this update. :(((

> * I see you are passing parameters by their name for init file. Did you check carefully so there are no typos?

The purpose of pass all named parameters as named is to avoid mistakes on updates related to sequence of parameters. With named values calls are now safe on parameters order misses.

I did this update as accurate as I can compare everything. As far as tests are passing - syntactically these calls are correct. Of course, there MAY BE some mistakes, I'll fix them on occurence.

telebot/types.py
Badiboy: > What is the purpose of kwargs when we don't use it?

IMO **kwargs saves library from updates of API. If there appears new fields - they will be autoparsed and passed to init. If there will be no dummy **kwargs to accept them - init call will fail with incorrect number of aguments. Isn't it?

We (you :) ) created these **kwargs long ago when invented this new method of calling init: I just allign some methods to the same signature as others.

telebot/types.py
Badiboy: Thanks, fixed.

coder2020official:
I think that is it

Badiboy:
Thank you for review!

Badiboy:
@coder2020official 
1. What about version? 4.15.0 or 5.0.0? :) 
2. Do we have to make anything else? Or I can start testing update on my systems?

coder2020official:
I wouldn't rush tbh. I am planning to:
- Update docstrings
- Update docs
- And try out black formatting

Anyways, I think we have to wait for some time. We don't want surprises, do we? And regarding version number - we need something more than just an API update for version 5. So I guess it would be better to stick with 4.15.0

Badiboy:
> Anyways, I think we have to wait for some time. We don't want surprises, do we?

Tests are part of activities avoiding surprises ))

coder2020official:
The project is not fully covered by tests which makes it harder for us to rely on them

PR #2112, opened 2024-01-05
Black implementation test

coder2020official:
@Badiboy this is how approximately black looks like

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -23,7 +23,7 @@ import string
 
 import ssl
 
-logger = logging.getLogger('TeleBot')
+logger = logging.getLogger("TeleBot")
 
 formatter = logging.Formatter(
     '%(asctime)s (%(filename)s:%(lineno)d %(threadName)s) %(levelname)s - %(name)s: "%(message)s"'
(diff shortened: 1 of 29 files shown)


Badiboy:
I PERSONALLY don't like this type of formatting. 

image

It gives nothing valuable but lot of lines. It's just one file: you got 1400 lines for nothing. It's valuable for ones who are paid for lines of code :)

coder2020official:
Idk, since you like changing formatting frequently, I thought you would like an automated system for that

coder2020official:
I will see whether I can change something in settings though to avoid some of the changes

Badiboy:
> Idk, since you like changing formatting frequently

I don't like ))) I just make some updates to formatting while making valuable updates ))

coder2020official:
I just tried yapf with a style of facebook, here are some screenshots if you are interested:
image
image
image

This one is still similar to black in some ways, but there are differences. Thoughts? We still have to try to obey PEP-8?

coder2020official:
We can technically ignore apihelper-like files(also asyncio_helper); Maybe in the future I will even try to get rid of these files due to a mess over there

Badiboy:
May be just live things as is? 😂😂😂

coder2020official:
Well, this is not my first time caring about PEP-8 - it is better if we at least try to follow it to some extent.

Badiboy:
I'm not in favour of carry about anything "because of" - if that does not bring any value. I do not see what value black formatting brings except "now we hava black formatting"...

coder2020official:
It indeed brings the value of cleaner code

Badiboy:
"Cleaner code" is not a value. It's "cleaner code bacause of cleaner code".

coder2020official:
Ok, if it does not harm project in any way - why not accept the new formatting?

Badiboy:
"Add value" is the good reason of bring changes.
"Why not" is the bad reason of bring changes.

coder2020official:
Well then I can argue that by improving formatting, we follow suggested PEP-8

coder2020official:
And this is a better practise. The reason is "adding a value", not "why not"

Badiboy:
"we follow suggested PEP-8" and it will give the project.... What? "Clear code"?

Badiboy:
"this is a better practise" - do you follow all "better practices" in your life because these are better practices?

coder2020official:
You should try to follow best practices as far as you can. They are created not for just existence but to help you understand your code faster

Cub11k:
> I PERSONALLY don't like this type of formatting. 
> 
> image
> 
> It gives nothing valuable but lot of lines. It's just one file: you got 1400 lines for nothing. It's valuable for ones who are paid for lines of code :)

Consistency in the code formatting for better readability is kinda worth it, provided no one really cares about the number of lines and the size of the py-files (at least in my opinion)

PR #2115, opened 2024-01-06
Doc translate

l27001:
Translated description of the get_user_chat_boosts method for sync and async versions


coder2020official:
Thank you

PR #2117, opened 2024-01-06
Doc update

l27001:
Translated `add_sticker_to_set`, `message_reaction_handler`, `message_reaction_count_handler`, `register_message_reaction_count_handler`, `register_message_reaction_handler`, `set_message_reaction` to russian

Also description of `set_message_reaction` says:
> The bot must be an administrator in the chat for this to work and must have the appropriate admin rights.

But this is incorrect. Bot doesn't require any admin rights to use this method. So I removed that part from the description


coder2020official:
You cannot edit english version directly. English version is dependent on docstrings in the package files. So you will have to edit async_telebot.py file descriptions too, then re-generate documentation

PR #2123, opened 2024-01-10
Fix disable_web_page_preview in LinkPreviewOptions

Badiboy:
## Description
Fix disable_web_page_preview to is_disabled in LinkPreviewOptions


coder2020official (approved):

coder2020official:
I will do a PR to fix some things soon, then I think we can release the version

PR #2124, opened 2024-01-12
Changes in deprecated parameters for async

coder2020official:
Separated allow_sending_without_reply and reply_to_message_id into 2 different if-elses for each methods;
Planning to add hcite and mcite

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -1939,10 +1939,10 @@ class TeleBot:
     def send_dice(
             self, chat_id: Union[int, str],
             emoji: Optional[str]=None, disable_notification: Optional[bool]=None, 
-            reply_markup: Optional[REPLY_MARKUP_TYPES]=None, 
             reply_to_message_id: Optional[int]=None,          # deprecated, for backward compatibility
-            allow_sending_without_reply: Optional[bool]=None, # deprecated, for backward compatibility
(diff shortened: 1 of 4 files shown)


coder2020official:
Btw, Badiboy, I noticed one thing: the order of parameters in a function is changed for deprecation-related methods. I think it will cause some issues after release. So for example, here is send_document:
image
I know passing arguments by their position is never recommended, however, don't you think this will cause issues after the release?

Badiboy:
> I know passing arguments by their position is never recommended

As for me, I don't think we should care about ones, who passes named parameters by position...

For this case it was not necessary: I changed the place of parameters because it helped me to update lot of functions. If you hardly care about all species - we may rollback that. We mostly means "you", because have neither time nor willing...

coder2020official:
I will find time to do that then. Don't wanna see reports regarding crashes after release

coder2020official:
I think it is time to check the PR

coder2020official:
@Badiboy you are here and aware that I am done with PR right? Just asking

Badiboy:
@coder2020official Sorry, I missed your request. Checking.

Badiboy (approved):

coder2020official:
I will go through one last time and notify you when everything is ready for release - is that fine?

Badiboy:
yep

PR #2127, opened 2024-01-14
Fix list objects in methods

coder2020official:


coder2020official:
@badiboy you think this doesn't require an extra release?

Badiboy:
I started testing the release and also found this bug. ) And other more sensitive. 

Yes, it will require a bugfix release.

coder2020official:
Could you carry out more tests and release a new version then?

Badiboy:
Yep, I will.

PR #2129, opened 2024-01-14
Bump version to 4.15.1 (bugfix)

Badiboy:
## Description
Bump version to 4.15.1 (bugfix release)


Badiboy:
@coder2020official I'll push it to pypi only for now, then will go for more testing. If succeed - I will publish elsewhere.

PR #2138, opened 2024-01-21
Fix ChatBoostSource issues

Badiboy:
## Description
Fix ChatBoostSource issues

1. "source" is temporary made optional
2. source types are parsed correctly

### DIFF
--- a/telebot/types.py
+++ b/telebot/types.py
@@ -9146,7 +9146,7 @@ class ChatBoost(JsonDeserializable):
     :param source: Optional. Source of the added boost (made Optional for now due to API error)
     :type source: :class:`ChatBoostSource`
 
-    :return: Instance of the class (made Optional for now due to API error)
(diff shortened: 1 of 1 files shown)


coder2020official:
LGTM, but let's wait for the answer from bot API dev

Badiboy:
> but let's wait for the answer from bot API dev

By making source "Optional" we do not break anything, just fix the current issue for users. When we got answer - we cann roll it back, it's not a problem :)

coder2020official:
https://github.com/tdlib/telegram-bot-api/blob/ade0841d41b7126c2d908475291a5e1cd8a1b905/telegram-bot-api/Client.cpp#L3345
It is not optional; I doubt that it is an API bug
We have to investigate the issue further

Badiboy:
It's wrong to wait more. 

@coder2020official When something will change - will make an update.

PR #2156, opened 2024-02-03
Aiohttp version updated; Python 3.12 support

coder2020official:


coder2020official:
@Badiboy looks like we have to update our installer :(
https://github.com/python/cpython/issues/95299

Badiboy:
Add setuptools to requirements?

coder2020official:
Is it a good idea?

coder2020official:
https://peps.python.org/pep-0518/
https://peps.python.org/pep-0517/
We need changes, specifically we need pyproject.toml and other things; we should remove setup.py to follow future releases

PR #2157, opened 2024-02-04
Testing out pyproject.toml instead of legacy setup.py

coder2020official:


coder2020official:
@Badiboy take a look

Badiboy:
@coder2020official I don't know anything about setups: never interested. If it really works - thus it's fine...

coder2020official:
Then merge; Should work I guess

coder2020official:
Oh btw... Version.py is now useless, I didn't delete it; I think some users get version from version.py, so let's leave it; So you will have to change version in both files

PR #2171, opened 2024-02-16
Fix Python versions 3.8-3.12

Badiboy:
## Description
Fix Python versions 3.8-3.12


coder2020official:
Did we drop support for python 3.7?

Badiboy:
https://devguide.python.org/versions/

It's ended.
image

Issue #2182, opened 2024-02-17
MenuButton implementation


KingRotiv:
When using the specific menu button, shouldn't it already fill in the type? I don't understand why even passing the specific class I need to pass the type.

https://github.com/eternnoir/pyTelegramBotAPI/blob/db4899a6b92b690ecb26ca28219de64eeac7deae/telebot/types.py#L7321-L7430

Badiboy:
Loks reasonable...

PR #2183, opened 2024-02-18
MenuButton set default types

Badiboy:
## Description
MenuButton set default types

No need to pass "type" if used specific MenuButtonX class.


Badiboy:
@coder2020official 
I see this issue reasonable: https://github.com/eternnoir/pyTelegramBotAPI/issues/2182
I don't like much the given solution, but I see no other way without breaking compatibility.

coder2020official (approved):

PR #2190, opened 2024-02-27
fix(hint): add_sticker_to_set -> sticker

sudoskys:
## Description
Method hint does not match the actual operation

## Describe your tests
Only the hint is fixed, the implementation remains unchanged

Python version: ALL

OS: ALL

## Checklist:


Badiboy:
Thank you.

PR #2209, opened 2024-04-02
Sticker set initialization fix

evd0kim:
## Description

Fixed 

```
 __init__() missing 2 required positional arguments: 'is_animated' and 'is_video'
```

by re-arranging arguments and assigning default values.

Tested on frequently updating stickers https://t.me/addstickers/hcn_by_SatoshiSaysBot

### DIFF
--- a/telebot/types.py
+++ b/telebot/types.py
@@ -6308,7 +6308,7 @@ class StickerSet(JsonDeserializable):
             obj['thumbnail'] = None
         return cls(**obj)
 
-    def __init__(self, name, title, sticker_type, is_animated, is_video, stickers, thumbnail=None, **kwargs):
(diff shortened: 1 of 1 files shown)


coder2020official:
API update caused this; @Badiboy

Badiboy:
Will be implemented in 7.2 update. When we find the way how to save backward compatibility.

evd0kim:
> Will be implemented in 7.2 update. When we find the way how to preserve backward compatibility.

I am just curious what backward compatibility is meant here

Badiboy:
If you simply change signature of existing library functions - all software that used that function before will be f***ed. :)

Badiboy:
@evd0kim Your solution was correct: we identified that this class is never created by users. 

Applied with some additions here: https://github.com/eternnoir/pyTelegramBotAPI/pull/2217

PR #2217, opened 2024-04-05
Bot API 7.2 u1

Badiboy:
- Removed the fields is_animated and is_video from the class [StickerSet](https://core.telegram.org/bots/api#stickerset).
- Removed the parameter sticker_format from the method [createNewStickerSet](https://core.telegram.org/bots/api#createnewstickerset).

Aded handlers to self.typed_middleware_handlers.
Minor deprecation fixes.

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -5890,7 +5890,7 @@ class TeleBot:
             stickers = [types.InputSticker(sticker=stickers, emoji_list=emojis, mask_position=mask_position)]
 
         if sticker_format:
-            logger.warning('The parameter "sticker_format" is deprecated since Bot API 7.2+. Stickers can now be mixed')
+            logger.warning('The parameter "sticker_format" is deprecated since Bot API 7.2. Stickers can now be mixed')
 
         return apihelper.create_new_sticker_set(
             self.token, user_id, name, title, stickers, sticker_type=sticker_type, needs_repainting=needs_repainting)
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -213,6 +213,7 @@ class TeleBot:
         # handlers
         self.exception_handler = exception_handler
         self.update_listener = []
+
         self.message_handlers = []
         self.edited_message_handlers = []
         self.channel_post_handlers = []
@@ -247,6 +248,8 @@ class TeleBot:
                 'edited_message': [],
                 'channel_post': [],
                 'edited_channel_post': [],
+                'message_reaction': [],
+                'message_reaction_count': [],
                 'inline_query': [],
                 'chosen_inline_result': [],
                 'callback_query': [],
@@ -256,7 +259,13 @@ class TeleBot:
                 'poll_answer': [],
                 'my_chat_member': [],
                 'chat_member': [],
-                'chat_join_request': []
+                'chat_join_request': [],
+                'chat_boost': [],
+                'removed_chat_boost': [],
+                'business_connection': [],
+                'business_message': [],
+                'edited_business_message': [],
+                'deleted_business_messages': [],
             }
             self.default_middleware_handlers = []
         if apihelper.ENABLE_MIDDLEWARE and use_class_middlewares:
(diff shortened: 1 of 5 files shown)


telebot/__init__.py
```diff
@@ -5880,9 +5889,11 @@ def create_new_sticker_set(
                 raise ValueError('You must pass at least one sticker')
             stickers = [types.InputSticker(sticker=stickers, emoji_list=emojis, mask_position=mask_position)]
 
+        if sticker_format:
+            logger.warning('The parameter "sticker_format" is deprecated.')
```
coder2020official, line 5893: I'd clarify why: 
```suggestion
            logger.warning('The parameter "sticker_format" is deprecated since Bot API 7.2+. Stickers can now be mixed')
```

telebot/async_telebot.py
```diff
@@ -7208,10 +7208,11 @@ async def create_new_sticker_set(
                 raise ValueError('You must pass at least one sticker')
             stickers = [types.InputSticker(sticker=stickers, emoji_list=emojis, mask_position=mask_position)]
             
+        if sticker_format:
+            logger.warning('The parameter "sticker_format" is deprecated.')
```
coder2020official, line 7212:
```suggestion
            logger.warning('The parameter "sticker_format" is deprecated since Bot API 7.2+. Stickers can now be mixed.')
```

telebot/types.py
```diff
@@ -6428,15 +6427,20 @@ def de_json(cls, json_string):
             obj['thumbnail'] = None
         return cls(**obj)
 
-    def __init__(self, name, title, sticker_type, is_animated, is_video, stickers, thumbnail=None, **kwargs):
+    def __init__(self, name, title, sticker_type, is_animated=None, is_video=None, stickers=None, thumbnail=None, **kwargs):
```
coder2020official, line 6430: hmm.. setting stickers to None. Hopefully this will not get us problems in future?
Why not change order of params, as the class represents the sent sticker;

telebot/types.py
```diff
@@ -6428,15 +6427,20 @@ def de_json(cls, json_string):
...
         self.name: str = name
         self.title: str = title
         self.sticker_type: str = sticker_type
-        self.is_animated: bool = is_animated
-        self.is_video: bool = is_video
         self.stickers: List[Sticker] = stickers
         self.thumbnail: PhotoSize = thumbnail
 
+        if is_animated is not None:
```
coder2020official, line 6437: They will always be None from now on, meaning that they will always get deprecation warnings.
Instead, I propose turning them into property and showing deprecation warning?

coder2020official (changes requested):

telebot/types.py
Badiboy: We cannot change the order of parameters, it will break code which passed is_animated/is_video, i.e. **all code** who used this class...

telebot/types.py
Badiboy: None will not trigger the warning. NOT NONE will: "NOT NONE" means user use old-style call signature.

telebot/types.py
coder2020official: It is JsonDeserializable and just represents an object sent by API
So it is not a breaking change.

telebot/types.py
coder2020official: hmm true, wasn't attentive
Anyways, since the class represents an object from API, is_animated and is_video parameters are not passed. So my suggestion about totally removing these params but making them remain as properties with deprecation warning is good?

coder2020official (changes requested):

telebot/types.py
Badiboy: Looks you are right.

telebot/types.py
Badiboy: I did not checked if the class is created by user or not. If it is just deserializable - this is the right way. I'll update.

Badiboy: @coder2020official both properties will return TRUE?

Badiboy: Both properties will return FALSE?

Badiboy:
@coder2020official Udated.

coder2020official:
LGTM now. Did you try testing?

Issue #2232, opened 2024-04-19
Globally set parse_mode doesn't apply to InputMedia class


baatochan:
1. What version of pyTelegramBotAPI are you using?
 4.16.1
2. What OS are you using?
Manjaro Linux
3. What version of python are you using?
3.11.8

I have set the parse mode for a bot globally using `bot.parse_mode = "MarkdownV2"`. It is applied to every message I send using `send_photo()`, `send_video()` or `send_animation()`. However when I use `send_media_group()` with a list of `InputMediaPhoto` or `InputMediaVideo` I have to manually specify the parse_mode for every `InputMedia` object. Is it an expected behavior?

Badiboy:
Looks reasonable.

Badiboy:
https://github.com/eternnoir/pyTelegramBotAPI/pull/2235

Issue #2227, opened 2024-04-18
An unmasked bot token is registered in the exclusion of connection to the Telegram API


orenlab:
Please answer these questions before submitting your issue. Thanks!

1. What version of pyTelegramBotAPI are you using?
4.17.0

3. What OS are you using?
Alpine Linux, Ubuntu on Docker container (see full [Dockerfile](https://github.com/orenlab/pytmbot/blob/master/Dockerfile))

4. What version of python are you using?
3.12.3 on Docker container

If an exception occurs while trying to connect to the Telegram API, an unmasked token for the bot will be printed to the standard output.

Example (full traceback):

```
Traceback (most recent call last):
  File "/venv/lib/python3.12/site-packages/urllib3/connectionpool.py", line 537, in _make_request
    response = conn.getresponse()
               ^^^^^^^^^^^^^^^^^^
  File "/venv/lib/python3.12/site-packages/urllib3/connection.py", line 466, in getresponse
    httplib_response = super().getresponse()
                       ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/http/client.py", line 1428, in getresponse
    response.begin()
  File "/usr/local/lib/python3.12/http/client.py", line 331, in begin
    version, status, reason = self._read_status()
                              ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/http/client.py", line 292, in _read_status
    line = str(self.fp.readline(_MAXLINE + 1), "iso-8859-1")
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/socket.py", line 707, in readinto
    return self._sock.recv_into(b)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/ssl.py", line 1252, in recv_into
    return self.read(nbytes, buffer)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/ssl.py", line 1104, in read
    return self._sslobj.read(len, buffer)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TimeoutError: The read operation timed out

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/venv/lib/python3.12/site-packages/requests/adapters.py", line 486, in send
    resp = conn.urlopen(
           ^^^^^^^^^^^^^
[...]

PR #2235, opened 2024-04-21
Some issue fixes

Badiboy:
## Description

- Pass default parse mode to Media items:
https://github.com/eternnoir/pyTelegramBotAPI/issues/2232
- Hide token in exception messages:
https://github.com/eternnoir/pyTelegramBotAPI/issues/2227

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -1030,6 +1030,11 @@ class TeleBot:
         self.event_observer.schedule(self.event_handler, path, recursive=True)
         self.event_observer.start()
 
+    def __hide_token(self, message: str) -> str:
+        if self.token in message:
+            code = self.token.split(':')[1]
+            return message.replace(code, "*" * len(code))
+
     def infinity_polling(self, timeout: Optional[int]=20, skip_pending: Optional[bool]=False, long_polling_timeout: Optional[int]=20,
                          logger_level: Optional[int]=logging.ERROR, allowed_updates: Optional[List[str]]=None,
                          restart_on_change: Optional[bool]=False, path_to_watch: Optional[str]=None, *args, **kwargs):
@@ -1083,15 +1088,9 @@ class TeleBot:
                              *args, **kwargs)
             except Exception as e:
                 if logger_level and logger_level >= logging.ERROR:
-                    e = str(e)
-                    if self.token in e:
-                        e = e.replace(self.token, 'TOKEN')
-                    logger.error("Infinity polling exception: %s", str(e))
+                    logger.error("Infinity polling exception: %s", self.__hide_token(str(e)))
                 if logger_level and logger_level >= logging.DEBUG:
-                    e = traceback.format_exc()
-                    if self.token in e:
-                        e = e.replace(self.token, 'TOKEN')
-                    logger.error("Exception traceback:\n%s", e)
+                    logger.error("Exception traceback:\n%s", self.__hide_token(traceback.format_exc()))
                 time.sleep(3)
                 continue
             if logger_level and logger_level >= logging.INFO:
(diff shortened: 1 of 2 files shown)


Badiboy:
@coder2020official Please, take a look. It's short )

coder2020official:
Is showing token a big bug though? Isn't it necessary for debug purposes, etc?

coder2020official:
If you really want to hide the token - I propose adding some checks for token to contain colon; Then, add a function to mask the letter part of the token(the one after colon); and show this format of token

coder2020official:
So, when creating a TeleBot instance, check for ":" in token; raise error if not present
Then, in any exception, call a function to replace after-colon part with asterisk;

Badiboy:
> Is showing token a big bug though? Isn't it necessary for debug purposes, etc?

Big bug - no. Necessary for debug - also not. 
Token is more or less sencitive thing, thus showing it in logs may be harmful in some cases.

Badiboy:
> If you really want to hide the token - I propose adding some checks for token to contain colon; Then, add a function to mask the letter part of the token(the one after colon); and show this format of token

What the reason to do this processing? What additional value will be given by hiding the letter part instead of replace the token as a whole?

coder2020official:
- Use asterisk (if I understood your question directly)
- The first part shows Bot ID, which might be necessary for debugging

Badiboy:
> * Use asterisk (if I understood your question correctly)
> * The first part shows Bot ID, which might be necessary for debugging

+

Badiboy:
image

@coder2020official Should it be await-ed?

coder2020official:
no

telebot/async_telebot.py
```diff
@@ -367,9 +367,15 @@ async def infinity_polling(self, timeout: Optional[int]=20, skip_pending: Option
...
                 if logger_level and logger_level >= logging.ERROR:
+                    e = str(e)
+                    if self.token in e:
+                        e = e.replace(self.token, 'TOKEN')
                     logger.error("Infinity polling exception: %s", str(e))
                 if logger_level and logger_level >= logging.DEBUG:
-                    logger.error("Exception traceback:\n%s", traceback.format_exc())
+                    e = traceback.format_exc()
+                    if self.token in e:
+                        e = e.replace(self.token, 'TOKEN')
```
coder2020official, line 377: mmm?

telebot/async_telebot.py
```diff
@@ -367,9 +367,15 @@ async def infinity_polling(self, timeout: Optional[int]=20, skip_pending: Option
...
+                    logger.error("Exception traceback:\n%s", e)
```
coder2020official, line 378: Did you forget about this part?

coder2020official (changes requested):

telebot/async_telebot.py
Badiboy: Ups...

telebot/__init__.py
```diff
@@ -8110,11 +8127,17 @@ def register_business_message_handler(self,
...
+
+        :param regexp: Regular expression
+        :type regexp: :obj:`str`
+
         :param func: Function executed as a filter
         :type func: :obj:`function`
 
-        :param pass_bot: True if you need to pass TeleBot instance to handler(useful for separating handlers into different files)
-        :type pass_bot: :obj:`bool`
+        :param content_types: Supported message content types. Must be a list. Defaults to ['text'].
```
coder2020official, line 8139: does it?...
we are not setting content_types

coder2020official (approved):
I am not sure on some parts, as I think some parts may not throw token error; Anyways, it is not bad to still have replacing there; So I think it looks good now

telebot/__init__.py
Badiboy: image

I just align description with parameters.

coder2020official: Ah ok, seems like we need another full PR with all content_types description fixes + default values. you can merge

Badiboy:
🤝

PR #2240, opened 2024-04-24
Update to resolve Issue with outdated `_convert_entities` function in `copy_message` method

YouKnow-sys:
## Description
I identified an issue with the outdated function `_convert_entities` used in both the sync and async versions of the `copy_message` method. This function was not functioning correctly, resulting in invalid entities. While all other methods were using a more updated method for converting entities, this particular method was still relying on `_convert_entities`. 

To address this issue, I have removed `_convert_entities` and replaced it with more better `json.dumps(types.MessageEntity.to_list_of_dicts(...))`. This change ensures that the entities are properly converted and are valid

Python version:
3.12

OS:
windows 10

## Checklist:

### DIFF
--- a/telebot/apihelper.py
+++ b/telebot/apihelper.py
@@ -432,7 +432,7 @@ def copy_message(token, chat_id, from_chat_id, message_id, caption=None, parse_m
     if parse_mode:
         payload['parse_mode'] = parse_mode
     if caption_entities is not None:
-        payload['caption_entities'] = _convert_entites(caption_entities)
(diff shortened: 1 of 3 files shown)


Badiboy:
Looks reasonable, thank you.

Badiboy:
image

Could you please enchance MessageEntity.to_list_of_dicts with _convert_entities checks? 
1. None or empty => None
2. If not serializable => entities
3. Else - make list.

Badiboy:
LGTM. Thank you.

PR #2286, opened 2024-05-30
Resolve the issue with inline queries

itisrasoul:
## Description
When the 'allowed_updates' parameter is not specified, it just uses some settings that does not allow updating the inline queries. So based on the docstring from the 'infinity_polling' method, we have to specify an empty list for it to allow any update, including the 'inline_queries' update.

## Describe your tests
Just tested the example file.

Python version:
3.11
OS:
Windows 10
## Checklist:

### DIFF
--- a/examples/inline_example.py
+++ b/examples/inline_example.py
@@ -61,7 +61,7 @@ def default_query(inline_query):
 
 
 def main_loop():
-    bot.infinity_polling()
+    bot.infinity_polling(allowed_updates=[])
     while 1:
         time.sleep(3)
 


Badiboy:
> it just uses some settings that does not allow updating the inline queries

It uses the default TELEGRAM behaviour. 
https://core.telegram.org/bots/api#getupdates
And it's correct to follow Telegram by default.

Badiboy:
For this example it looks reasonable to obviously set allowed_updates to what is needed for example. Inluding _inline_queries_. Or even mention all... AFAIK there is a constant with list of actions...

Badiboy:
utils.update_types

And this also abviously mentioned in infinity_polling description...

itisrasoul:
Well as a beginner it took me some time to find that out, and one of the main usages for the 'example' section is to show beginners how some features work. So yeah, I guess it's better to just set `allowed_updates` to the needed value, rather than following the default behavior. 
Plus, I came across this in the docstring of `inifinity_polling`:

image

I don't know what it means by "the previous setting".

coder2020official:
Previous setting according to API.
A beginner would not play around with allowed_updates list.
Editing one example to just have specific update_types would mean editing every other example which is ridiculous.

Badiboy:
> Editing one example to just have specific update_types would mean editing every other example which is ridiculous.

I propose to set "utils.update_types" there.

coder2020official:
Then every example needs to be edited

Badiboy:
May be this have a reason. )

coder2020official:
Waste of time. By default, all needed updates are provided by Telegram. Only those who do not read the documentation mess things up and complain

PR #2295, opened 2024-06-07
really drop python<=3.7 support

kloczek:
Filter all code over `pyupgrade --py38-plus`.

## Checklist:


Badiboy:
@coder2020official As always I see absolutely no reasons in this update. It gives no value except lot of changes and lot of pain to users with old python. No FORCE DROP of python 3.7 was done, just drop CHECKING of python 3.7 compatibility for futute updates.

Badiboy:
I reject this PR.

PR #2313, opened 2024-06-22
Added parameter "timeout" to edit_xxx functions

Badiboy:
## Description
Added parameter "timeout" to edit_xxx functions


coder2020official:
maybe release new version first

PR #2321, opened 2024-06-23
StatesV2: Support for topics, multibots, and business messages

coder2020official:
~~Only sync memorystorage working now~~


coder2020official:
@Badiboy I might need your opinion for further development. Have you used states before I ask you a question?

Badiboy:
@coder2020official I never used States :(

coder2020official:
Finished for sync - only left thing is to add a "bridge" between old and new formats of states to ease switching between these two.

telebot/async_telebot.py
```diff
@@ -183,6 +186,14 @@ def __init__(self, token: str, parse_mode: Optional[str]=None, offset: Optional[
         self.middlewares = []
 
         self._user = None # set during polling
+        self.bot_id = None
+
+        if token_check:
+            result = apihelper.get_me(token)
```
coder2020official, line 192: @badiboy do you think this is a good idea

coder2020official:
Checklist:

- [ ] Rewrite sync pickle storage to use decorators for locks to follow DRY
- [ ] Add a method for redis storage to migrate from old to new storage type
- [ ] test with old code example to ensure old code compatibility
- [ ] Compare old methods with new methods for all storages to check for return types
- [ ] Check typehints
- [ ] Add comments to needed functions & classes to provide an overview and usage 
- [ ] Update example to ensure better state experience
- [ ] Discuss the synchronous getMe call with @ badiboy for async
- [ ] Discuss the file structure implementation and file names(e.g states, aio & sync folder names, maybe better asyncio?)

coder2020official:
Pickle storage is the worst out of all. I had some bugs when simultaneously testing my states code. 
I will investigate the issue and add a warning that pickle should NOT be used for production. Maybe for testing to a small audience.

coder2020official:
<Breaking changes will be here>
1. state="*" filter will check for existing state, instead of returning True in StateFilter, which is useless; Not yet implemented.

coder2020official:
@Badiboy i need your opinion on something I addressed before. Check the comment above.

telebot/async_telebot.py
coder2020official: So basically due to sync __init__, I had to call getMe from apihelper to obtain types.User due to token_check being true(new param)

Badiboy: > So basically due to sync **init**, I had to call getMe from apihelper to obtain types.User due to token_check being true(new param)

Looks a bit ugly, but functional.

coder2020official:
@Badiboy I am nearly done with this PR.
I have provided all breaking changes which I believe are all reasonable; If they were not - I provided a way to fix backward compatibility(e.g migrate_format for Redis storage to avoid data loss)

Now, I would like you to review:
- Folder & file names. I have moved states to a folder called states, with aio and sync folders. Do you think these names are fine? Maybe I should rename "aio" to "asyncio"?
- the whole code :) (or at least the main files - init.py, async_telebot.py, etc to ensure backward compatibility, etc)

coder2020official:
All that is left is cleaning up the code, which I will do after you proceed with the review

coder2020official:
@badiboy I have just noticed some unexpected behaviour. It seems like the getMe call during init of the class is slowing down and resulting in connection timeout.
Could you please test this out? Take echo bot code with this version of library installed, and try re-starting to see for any side effects.

I think placing a request in init is not a good idea after all. Do you have any ideas on where I can properly get the user object of the bot?(((
Polling is one place, but to guarantee work of states(passing bot_id automatically), I need to ensure this object also exists for webhooks

coder2020official:
- Changed token_check -> validate_token, which will just validate token by inspecting the string instead of calling getMe(wasn't a good idea after all); Drawbacks: subject to change in future. I am treating the first part of token as bot id, which is unspecified and may be changed in future, however, a number should still work for states.

coder2020official:
@Badiboy I am fully done, need your review

Badiboy:
@coder2020official I saw the request but I'm not sure when I'll be able to pass all of this:

image

Especially as far as I never used stated.

coder2020official:
You can review only crucial files(not news ones, etc), e.g init, async_telebot.py, etc
All new files or storages do not require a review

coder2020official:
Also I have addressed all breaking changes above

telebot/__init__.py
```diff
@@ -185,6 +191,12 @@ def __init__(
         self.allow_sending_without_reply = allow_sending_without_reply
         self.webhook_listener = None
         self._user = None
+        self.bot_id: int = None
+
+        if validate_token:
+            util.validate_token(self.token)
+
+        self.bot_id = util.extract_bot_id(self.token) # subject to change in future, unspecified
```
Badiboy, line 199: May cause error if token is not in valid form. extract_bot_id should return NONE if extract fails.

telebot/__init__.py
```diff
@@ -1172,6 +1184,9 @@ def polling(self, non_stop: Optional[bool]=False, skip_pending: Optional[bool]=F
         if restart_on_change:
             self._setup_change_detector(path_to_watch)
 
+        if not self._user:
```
Badiboy, line 1187: We discussed that get_me should be optional.

telebot/__init__.py
```diff
@@ -6644,57 +6661,104 @@ def set_state(self, user_id: int, state: Union[int, str, State], chat_id: Option
...
+            business_connection_id=business_connection_id, message_thread_id=message_thread_id)
 
 
-    def reset_data(self, user_id: int, chat_id: Optional[int]=None):
+    def reset_data(self, user_id: int, chat_id: Optional[int]=None, 
+                     business_connection_id: Optional[str]=None,
+                     message_thread_id: Optional[int]=None, bot_id: Optional[int]=None) -> bool:
         """
-        Reset data for a user in chat.
+        Reset data for a user in chat: sets the 'data' fieldi to an empty dictionary.
```
Badiboy, line 6706: fieldi => field?

telebot/async_telebot.py
```diff
@@ -183,6 +189,13 @@ def __init__(self, token: str, parse_mode: Optional[str]=None, offset: Optional[
         self.middlewares = []
 
         self._user = None # set during polling
+        self.bot_id: int = None
+
+        if validate_token:
+            util.validate_token(self.token)
+        
+        self.bot_id: int = util.extract_bot_id(self.token) # subject to change, unspecified
```
Badiboy, line 197: May cause error if token is not in valid form. extract_bot_id should return NONE if extract fails.

telebot/async_telebot.py
```diff
@@ -424,7 +437,8 @@ async def _process_polling(self, non_stop: bool=False, interval: int=0, timeout:
             # show warning
             logger.warning("Setting non_stop to False will stop polling on API and system exceptions.")
 
-        self._user = await self.get_me()
+        if not self._user:
+            self._user = await self.get_me()
```
Badiboy, line 441: We discussed that get_me should be optional.

Badiboy:
@coder2020official I checked only main files. All related to states are on your responcibility )

telebot/__init__.py
coder2020official: But validate_token raises an error if token is not in valid form?

telebot/async_telebot.py
coder2020official: I will return this to the original version, where it is called by default?

telebot/__init__.py
coder2020official: Yep, my bad, for sync there's no need for this as self.user is a property that already does this

telebot/__init__.py
Badiboy: Yes. But only if validate_token is enabled. It's optional.

telebot/__init__.py
coder2020official: oh hm, the use case might be where validate_token is false, will fix

telebot/__init__.py
coder2020official: oh yes. Btw, you are fine with validate_token being True by default?

Badiboy: I saw it. I'm not happy, but as far it is not breaking and can be disabled - I can deal with it.

coder2020official:
Anything else left?

Badiboy:
I don't understand your last commits.

1.
image
bot_id is available only for ones who used validate_token? Independently if token is valid or not? As I noted before, I propose still trying to get bot_id, but just handle the case when it cannot be extracted. It's not too hadr.

Badiboy:
2.
You totally removed get_me from sync:
image
And make it every time in async
image

Instead of make it optional in both versions?

coder2020official:
I returned them to their original versions.
Removing in sync will still cause a request to be done, as self.user will be empty, and it is a property.

Before, async would always call a getme request

coder2020official:
For 1), ok will fix it

coder2020official:
fixed 1), addressed 2)

Badiboy:
LGTM, let's go )

coder2020official:
merge?

coder2020official:
@badiboy bot api update was released just now.
I propose first releasing bot API update, then merge this

coder2020official:
Need to change version to 4.23.0

Badiboy:
Have nothing against it. Push it here and merge.

coder2020official:
I am afk, so we will have to delay this a little 
No reason to rush anyways

coder2020official:
@badiboy done, merging since you approved.
Next version will need to be 4.23.0

coder2020official:
We will also need a special changelog for next version with changes specifically for states.

coder2020official:
**States**

Breaking changes

- state="*" filter now checks if state exists.
- get_data returns an empty dictionary if record does not exist
- State storage format completely changed: in order to avoid state loss, StateRedisStorage().migrate_format(bot_id) should be used (only for redis as it is supposed to be the only storage used for production)

Features of new states
- Support for business bots
- Support for topics
- Support for multibots

To automatically set business_message_id or message_thread_id, check for examples that use StateContext:
https://github.com/eternnoir/pyTelegramBotAPI/blob/master/examples/custom_states.py

Issue #2327, opened 2024-06-30
Async Telebot retries immediately when it runs into a bad gateway error, potentially flooding the API


All-The-Foxes:
I'm using pyTelegramBotAPI version 4.20 on Ubuntu Noble (24.04 LTS) with Python 3.12. When async Telebot receives a bad gateway response from Telegram, it will retry immediately and log an unhandled exception. My own code logs the bad gateway error and then sleeps for 5 seconds. During this sleep, several more error messages per second about this unhandled exception will continue to be logged by async Telebot if the server's connection remains unstable, meaning Telebot itself is trying to reconnect over and over and over again for polling without waiting first. This not only floods the API, but it also generates massive log files full of complete garbage and spams the logging output.

https://github.com/eternnoir/pyTelegramBotAPI/assets/116322192/2c67ae1c-7724-4a34-866f-109b12425379

coder2020official:
Yep, this was reported. Thanks again, I will fix it ASAP

coder2020official:
#2332 should fix it, it introduces an error interval which increases twice after each unsuccessful getupdates request, until it reaches its maximum of 60 seconds.
<img width="980" alt="image" src="https://github.com/eternnoir/pyTelegramBotAPI/assets/63740404/a11a2a24-04d7-4809-9d80-8385e7bca995">

Issue #2320, opened 2024-06-23
Failed working with visible_file_name in async


Badiboy:
https://github.com/eternnoir/pyTelegramBotAPI/discussions/2319#discussioncomment-9850643

Looks like some kind of bug. @coder2020official , could you please check?

coder2020official:
Sure, will check it later

coder2020official:
- [ ] Should add a check for presence of both InputFile and visible_file_name and show a warning;
- [ ] file_name should be added as an optional parameter in InputFile

PR #2332, opened 2024-07-03
Error intervals to stop flooding logs & set non_stop to true by default

coder2020official:
#2327 #2331 #2320

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -2598,6 +2598,10 @@ class TeleBot:
             logger.warning('The parameter "thumb" is deprecated. Use "thumbnail" instead.')
             thumbnail = thumb
 
+        if isinstance(document, types.InputFile) and visible_file_name:
+            # inputfile name ignored, warn
+            logger.warning('Cannot use both InputFile and visible_file_name. InputFile name will be ignored.')
+
         return types.Message.de_json(
             apihelper.send_data(
                 self.token, chat_id, document, 'document',
--- a/telebot/async_telebot.py
+++ b/telebot/async_telebot.py
@@ -452,10 +452,11 @@ class AsyncTeleBot:
                     if not handled:
                         logger.error('Unhandled exception (full traceback for debug level): %s', self.__hide_token(str(e)))
                         logger.debug(self.__hide_token(traceback.format_exc()))
+
+                    if non_stop:
                         error_interval = await self._handle_error_interval(error_interval)
                         
                     if non_stop or handled:
-                        #await asyncio.sleep(2) # used error_interval instead
                         continue
                     else:
                         return
@@ -464,6 +465,8 @@ class AsyncTeleBot:
                     if not handled:
                         logger.error('Unhandled exception (full traceback for debug level): %s', self.__hide_token(str(e)))
                         logger.debug(self.__hide_token(traceback.format_exc()))
+
+                    if non_stop:
                         error_interval = await self._handle_error_interval(error_interval)
 
                     if non_stop or handled:
@@ -475,6 +478,8 @@ class AsyncTeleBot:
                     if not handled:
                         logger.error('Unhandled exception (full traceback for debug level): %s', str(e))
                         logger.debug(traceback.format_exc())
+
+                    if non_stop:
                         error_interval = await self._handle_error_interval(error_interval)
 
                     if non_stop or handled:
--- a/telebot/async_telebot.py
+++ b/telebot/async_telebot.py
@@ -246,7 +246,7 @@ class AsyncTeleBot:
         self.event_observer.schedule(self.event_handler, path, recursive=True)
         self.event_observer.start()
 
-    async def polling(self, non_stop: bool=False, skip_pending=False, interval: int=0, timeout: int=20,
+    async def polling(self, non_stop: bool=True, skip_pending=False, interval: int=0, timeout: int=20,
             request_timeout: Optional[int]=None, allowed_updates: Optional[List[str]]=None,
             none_stop: Optional[bool]=None, restart_on_change: Optional[bool]=False, path_to_watch: Optional[str]=None):
         """
@@ -256,11 +256,6 @@ class AsyncTeleBot:
         Warning: Do not call this function more than once!
         
         Always gets updates.
-
-        .. note::
-
-            Set non_stop=True if you want your bot to continue receiving updates
-            if there is an error.
 
         .. note::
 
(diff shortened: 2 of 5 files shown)


coder2020official:
@badiboy will need your review on this one too
The InputFile bit - now if both inputfile and visible_file_name are present for send_document, visible_file_name will be used for file name.

telebot/__init__.py
```diff
@@ -2598,6 +2598,10 @@ def send_document(
             logger.warning('The parameter "thumb" is deprecated. Use "thumbnail" instead.')
             thumbnail = thumb
 
+        if isinstance(document, types.InputFile) and visible_file_name:
+            # inputfile name ignored, warn
+            logger.warning('Cannot use both InputFile and visible_file_name. InputFile name will be ignored.')
```
Badiboy, line 2603: `files[key] = (value.file_name, value.file)`

visible_file_name will be ignoread, isn't it?

telebot/async_telebot.py
```diff
@@ -446,9 +452,10 @@ async def _process_polling(self, non_stop: bool=False, interval: int=0, timeout:
                     if not handled:
                         logger.error('Unhandled exception (full traceback for debug level): %s', self.__hide_token(str(e)))
                         logger.debug(self.__hide_token(traceback.format_exc()))
+                        error_interval = await self._handle_error_interval(error_interval)
                         
                     if non_stop or handled:
-                        await asyncio.sleep(2)
+                        #await asyncio.sleep(2) # used error_interval instead
```
Badiboy, line 458: Changed behaviour!

In old version sleep(2) is for non_stop or **handled**. In new version it is for **not hadled** and **independently from non_stop**.

1. Adding for not handled is fine.
2. Removing from handled **is not fine**, because handled exceptions will cause now immediate re-run instead of 2 second delay.
3. Make sleep() for non_stop looks also not needed.

I propose:

```
                    if not handled:
                        logger.error('Unhandled exception (full traceback for debug level): %s', self.__hide_token(str(e)))
                        logger.debug(self.__hide_token(traceback.format_exc()))

                    if non_stop:
                        error_interval = await self._handle_error_interval(error_interval)

```

telebot/async_telebot.py
```diff
@@ -457,6 +464,7 @@ async def _process_polling(self, non_stop: bool=False, interval: int=0, timeout:
                     if not handled:
                         logger.error('Unhandled exception (full traceback for debug level): %s', self.__hide_token(str(e)))
                         logger.debug(self.__hide_token(traceback.format_exc()))
+                        error_interval = await self._handle_error_interval(error_interval)
```
Badiboy, line 467: The same here.

```
                    if not handled:
                        logger.error('Unhandled exception (full traceback for debug level): %s', self.__hide_token(str(e)))
                        logger.debug(self.__hide_token(traceback.format_exc()))

                    if non_stop:
                        error_interval = await self._handle_error_interval(error_interval)

                    if non_stop or handled:
                        continue
```

telebot/async_telebot.py
```diff
@@ -467,6 +475,7 @@ async def _process_polling(self, non_stop: bool=False, interval: int=0, timeout:
                     if not handled:
                         logger.error('Unhandled exception (full traceback for debug level): %s', str(e))
                         logger.debug(traceback.format_exc())
+                        error_interval = await self._handle_error_interval(error_interval)
```
Badiboy, line 478: And here.

telebot/async_telebot.py
```diff
@@ -4014,6 +4023,10 @@ async def send_document(
         if reply_parameters and (reply_parameters.allow_sending_without_reply is None):
             reply_parameters.allow_sending_without_reply = self.allow_sending_without_reply
 
+        if isinstance(document, types.InputFile) and visible_file_name:
+            # inputfile name ignored, warn
```
Badiboy, line 4027: `files[key] = (value.file_name, value.file)`

visible_file_name will be ignoread, isn't it?

telebot/asyncio_helper.py
```diff
@@ -130,6 +130,8 @@ def _prepare_data(params=None, files=None):
             if isinstance(f, tuple):
                 if len(f) == 2:
                     file_name, file = f
+                    if isinstance(file, types.InputFile):
```
Badiboy, line 133: Cannot verify this, never worked with InputFile.

telebot/types.py
```diff
@@ -7735,8 +7735,11 @@ class InputFile:
             InputFile(pathlib.Path('/path/to/file/file.txt'))
         )
     """
-    def __init__(self, file) -> None:
-        self._file, self.file_name = self._resolve_file(file)
+    def __init__(self, file: Union[str, IOBase, Path], file_name: Optional[str] = None):
+        self._file, self._file_name = self._resolve_file(file)
+        if file_name:
```
Badiboy, line 7740: Why not )

telebot/__init__.py
coder2020official: No, it returns a tuple and in apihelper, tuple value[0] is unchanged, thus ignoring the inputfile naming

telebot/async_telebot.py
coder2020official: so we still don't need sleep for handled? or should I add sleep to handled (2 seconds) back?

Badiboy: We need to sleep both for handled (as before) and not handled (it's reasonable).

But I propose only if non_stop: otherwise sleeping is useless, because polling will stop anyway.

coder2020official: wait, before, script slept 2 seconds for handled or non_stop. So, should I:
- Change your last condition to "if non_stop or handled"?
- Add another condition for handled, making the script sleep for 2 seconds?

coder2020official: this will not affect not handled: there will be no sleeping as you said, the script will be useless. currently we are talking about handled. Or, do you think the user should manually do that and not rely on internal behaviour?

coder2020official: I am talking about the need to have "if handled" condition. 
In previous version, as you know, sleep is triggered on handled or non_stop. After I copy your suggestion, sleep is not triggered for handled. is that fine?

Badiboy: > After I copy your suggestion, sleep is not triggered for handled. is that fine?

It is not. But if you copy **my version** - it will work both for handled and not handled.

Badiboy: > After I copy your suggestion, sleep is not triggered for handled. is that fine?

It is not. 

I don't know what you did, but if you'll take **my version** - it will work both for handled and not handled. Isn't it?

coder2020official: ah, sorry 
was sleepy a little, should work, yep

Badiboy:
image

PR #2344, opened 2024-07-11
new: add register_next_step_handler for asyncbot

hiddify-com:
## Description
create missing 
register_next_step_handler  in asyncbot using auto state

## Describe your tests
run step_example.py in async_telebot

Python version:

OS:

## Checklist:
- [*] I added/edited example on new feature/change (if exists)
- [*] My changes won't break backward compatibility
- [*] I made changes both for sync and async

### DIFF
--- a/examples/asynchronous_telebot/step_example.py
+++ b/examples/asynchronous_telebot/step_example.py
@@ -1,0 +1,74 @@
+# -*- coding: utf-8 -*-
+"""
+This Example will show you how to use register_next_step handler.
+"""
+
+import asyncio
+
+from telebot import asyncio_filters, types
+from telebot.async_telebot import AsyncTeleBot
+from telebot.asyncio_storage.memory_storage import StateMemoryStorage
+
+API_TOKEN = '<api_token>'
+
+bot = AsyncTeleBot(API_TOKEN, state_storage=StateMemoryStorage())
+
+
+
+
+# Handle '/start' and '/help'
+@bot.message_handler(commands=['help', 'start'])
+async def send_welcome(message):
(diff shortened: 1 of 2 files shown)


coder2020official:
No

hiddify-com:
why?

coder2020official:
1. Bad implementation 
2. No steps in async please

hiddify-com:
It is based on your recommendation that suggest to use state

It is really useful to have steps in async

The step in sync is buggy also. It does not process middlewares.

coder2020official:
Steps are an old, somewhat "deprecated"(though not yet and won't probably), so there's no point of improving them when there's a solution(states) which I am currently working on to even further improve

hiddify-com:
the problem of state is that we can not have more complex workflow
for example when we want to get 10 elements from user it is so hard to be implemented by state

coder2020official:
What do you even mean by "10 elements"?
10 inputs?
I am working on an easier solution, some class that will take care of this

hiddify-com:
yes 10 inputs from the user

great thanks

hiddify-com:
@coder2020official Thanks for your great job. It would be great if we can have a timeline about your solution

coder2020official:
Hi, my solution never came to existence, and I haven't had enough time recently either. I wouldn't mind re opening this.

I'm open to proposals, or I'll look at this issue again in December and see if there's any way I can simplify states logic.

Also, I have re implemented states to make them easier to use :)

coder2020official:
I'll open an issue

hiddify-com:
Thanks

PR #2345, opened 2024-07-11
chg: allow callback_query_handler to do not have func arg.

hiddify-com:
## Description
when we have a custom filter forcing using func is not acceptable in callback_query_handler so i have removed it

## Describe your tests
How did you test your change?

Python version:

OS:

## Checklist:
- [*] I added/edited example on new feature/change (if exists)
- [*] My changes won't break backward compatibility
- [*] I made changes both for sync and async

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -7700,7 +7700,7 @@ class TeleBot:
         self.add_chosen_inline_handler(handler_dict)
 
 
-    def callback_query_handler(self, func, **kwargs):
+    def callback_query_handler(self, func=None, **kwargs):
         """
         Handles new incoming callback query.
         As a parameter to the decorator function, it passes :class:`telebot.types.CallbackQuery` object.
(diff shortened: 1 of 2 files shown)


Badiboy:
> when we have a custom filter forcing using func is not acceptable in callback_query_handler so i have removed it

Very (extremely) breaf reasoning. We (at least I) require detailed description of:
1. What for
2. How it will be used (with examples)
3. Proof that it will not break the backward compatibility

coder2020official:
Hi, this is actually a useful PR. Let me talk instead:
1. Func should not be a required parameter. After implementing custom filters, many use them and hence don't rely on func for obvious reasons. In addition, some handlers already have func as optional. 
2. @bot.callback_query(data=...) use a filter with data key instead of lambda function
3. Making a parameter optional shouldn't break the compatibility. Those who already pass values to it won't suffer, and those already setting it to None won't either. Filters are checked(telling this from my memory) and filter is discarded if it is None.

Feel free to quote me

Badiboy:
No problem. If you (@coder2020official ) approve it - it's ok.

coder2020official:
merge @Badiboy

PR #2334, opened 2024-07-03
make json viewer more clear

2ei:
**make printing types understandable by using json.dumps, and ignore fields with None value, as Bot API did.**

- I tested it here
```python
@bot.message_handler(content_types=["text"])
async def on_text_message(m):
    print(m)
```

- This was the result before ( so annoying and not clear )
```json
[...]

PR #2350, opened 2024-07-21
make json viewer more clear (reopen)

z44d:
reopened #2334

### DIFF
--- a/telebot/types.py
+++ b/telebot/types.py
@@ -106,23 +106,20 @@ class JsonDeserializable(object):
             else {
                 attr: getattr(obj, attr)
                 for attr in filter(
-                    (lambda x: not x.startswith("_") and getattr(obj, x) is not None)
-                    if JSONDESERIALIZABLE_SKIP_NONE
-                    else (lambda x: not x.startswith("_")),
+                    lambda x: not x.startswith("_")
+                    and (
+                        getattr(obj, x) is not None or not JSONDESERIALIZABLE_SKIP_NONE
+                    ),
                     obj.__dict__,
                 )
             }
         )
         return (
-            json.dumps(
-                self, default=default, indent=2, ensure_ascii=False
-            )
+            json.dumps(self, default=default, indent=2, ensure_ascii=False)
             if JSONDESERIALIZABLE_PARSE_OUTPUT
             else str(
                 {
-                    x: default(y)
-                    if isinstance(y, JsonDeserializable)
-                    else y
+                    x: default(y) if isinstance(y, JsonDeserializable) else y
                     for x, y in self.__dict__.items()
                     if y is not None or not JSONDESERIALIZABLE_SKIP_NONE
                 }
--- a/telebot/types.py
+++ b/telebot/types.py
@@ -19,6 +19,8 @@ from telebot.formatting import apply_html_entities
 
 
 DISABLE_KEYLEN_ERROR = False
+JSONDESERIALIZABLE_PARSE_OUTPUT = False
+JSONDESERIALIZABLE_SKIP_NONE = True
 
 logger = logging.getLogger('TeleBot')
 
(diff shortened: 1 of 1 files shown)


coder2020official:
I propose:

- Storing json_string, with 'from' being changed to 'from_user'
- When printing an object, this json_string should be printed.

And then, we can add some JSON_FORMATTED = FALSE or something like this to types.

@Badiboy need your opinion. This way, users will be able to see a CLEAR response from Telegram.

Badiboy:
> This way, users will be able to see a CLEAR response from Telegram.

Debug mode have full and clear responces of the Telegram. 

> * Storing json_string, with 'from' being changed to 'from_user'

I have no willing to waste my and others server resources for storing json strings that I will never used before and will not use in future for the only purpose that one guy after several years of library existence occasionally decided to have pretty output. 

Except you will make it SO OPTIONAL that it will affect NOBODY except they obviously decided to use it.

coder2020official:
> I have no willing to waste my and others server resources for storing json strings

Makes sense

We still need changes to the output of the types though; Then, we can stick to the current implementation, but with pretty-printing optional.
I have personally encountered parts where outputting classes is SO annoying: 99% of values shown are None. I have to turn debug afterwards.
While debug is still needed, I see it reasonable to remove None values by default; And make pretty-printing optional by default.

Badiboy:
I have nothing against pretty output as I said from the very beginning which is: 
1. Optional
2. Not consuming resources for those who do not need it

Intermediate version that was discussed that allowed:
1. Optionally enable pretty output
2. Optionally disable NONEs
Looked fine for me.

coder2020official:
Is disabling Nones by default fine with you?

Badiboy:
Mmm... Let's be like that as far as Nones are really useless and it should be more or less clear that if you do not see property - it's empty.

coder2020official:
Well I don't see any use case for seeing Nones. They are useless as API does not provide them.
They just take up space, nothing useful out of this.

coder2020official:
Allright!
@2ei then could you please:

**Create 2 variables:**

1. JSONDESERIALIZABLE_PARSE_OUTPUT = False
2. JSONDESERIALIZABLE_SKIP_NONE = True

Use these to pretty print and output None values if necessary.

2ei:
@coder2020official Done

coder2020official:
@badiboy do you think parse indent is this necessary

Badiboy:
> parse indent

What's that?

telebot/types.py
```diff
@@ -97,12 +99,32 @@ def check_json(json_type, dict_copy = True):
...
+                    lambda x: not x.startswith("_")
+                    and (
+                        getattr(obj, x) is not None or not JSONDESERIALIZABLE_SKIP_NONE
+                    ),
+                    obj.__dict__,
+                )
+            }
+        )
+        return (
+            json.dumps(self, default=default, indent=2, ensure_ascii=False)
```
coder2020official, line 118: indent number here

coder2020official:
@2ei just implement it, won't hurt

coder2020official:
Thanks for the PR. Will test it out ASAP

telebot/types.py
```diff
@@ -97,12 +100,32 @@ def check_json(json_type, dict_copy = True):
...
             raise ValueError("json_type should be a json dict or string.")
 
-    def __str__(self):
-        d = {
-            x: y.__dict__ if hasattr(y, '__dict__') else y
-            for x, y in self.__dict__.items()
-        }
-        return str(d)
+    def __str__(self) -> str:
+        default = (
```
Badiboy, line 104: Frankly speaking, I do not understand anything in this pigeon talk. It takes too much time to decode.

Can this code be rewriten in adequate if / then / else mode? So ANYBODY who see it can understand what will be returned in what case?

2ei: OK

Badiboy:
@2ei Thank you.
LGTM if pass tests. )

coder2020official:
Screenshot 2024-08-11 at 4 55 16 PM
LGTM. Just one more think: @badiboy do you think it would be better if this PR also removes json field in the output? or do you think it will be necessary? It is just a repetition of the whole dict.

Badiboy:
Romoving json looks reasonable: no need to duplicate the same data.

2ei:
> Romoving json looks reasonable: no need to duplicate the same data.

what should i do in your opinion?, convert `json` parameter to `property` method or just make an condition in `JsonDeserializable` ?

coder2020official:
LGTM. What did we agree on json?

z44d:
> > Romoving json looks reasonable: no need to duplicate the same data.
> 
> what should i do in your opinion?, convert `json` parameter to `property` method or just make an condition in `JsonDeserializable` ?
> 
> Also i just noticed that there is no type hint for json parameter, on my way to add it too.

idk, i have 2 suggestions here, what is the best one for you @coder2020official ?

coder2020official:
@badiboy this has been on for a long time. do you think we should convert json to property?

PR #2353, opened 2024-07-27
Improved typehints for types.py #2351

coder2020official:
#2351

### DIFF
--- a/telebot/types.py
+++ b/telebot/types.py
@@ -5,7 +5,7 @@ from io import IOBase
 import logging
 import os
 from pathlib import Path
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, Optional, Union, Any, Tuple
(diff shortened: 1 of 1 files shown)


telebot/types.py
```diff
@@ -4209,6 +4214,7 @@ def to_dict(self):
             json_dict['is_flexible'] = self.is_flexible 
         return json_dict
 
+InputMessageContent = Union[InputTextMessageContent, InputLocationMessageContent, InputVenueMessageContent, InputContactMessageContent, InputInvoiceMessageContent]
```
coder2020official, line 4217: One notable change:
Since InputXXXMessageContent do not have a common class, I decided to declare this

Badiboy:
@coder2020official I cannot check it all ))) As far as hints are not significant, let's merge and fix later if any.

Issue #2349, opened 2024-07-21
Use deepcopy instead of shallow copy in _run_middlewares_and_handlers for middleware data 


coder2020official:
This will allow passing things to post_process in case handler does not want them.

PR #2354, opened 2024-07-27
Use deepcopy instead of a shallow copy for middlewares, #2349

coder2020official:
#2349

### DIFF
--- a/telebot/async_telebot.py
+++ b/telebot/async_telebot.py
@@ -32,7 +32,7 @@ REPLY_MARKUP_TYPES = Union[
 import string
 import random
 import ssl
-
+import copy
 
 """
 Module : telebot
@@ -558,7 +558,7 @@ class AsyncTeleBot:
                             logger.error("It is not allowed to pass data and values inside data to the handler. Check your handler: {}".format(handler['function']))
                             return
                     else:
-                        data_copy = data.copy()
+                        data_copy = copy.deepcopy(data)
                         for key in list(data_copy):
                             # remove data from data_copy if handler does not accept it
                             if key not in params:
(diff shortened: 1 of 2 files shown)


Badiboy:
I'm not sure that is the best way.

As far as I understand - you want this to allow middleware developer to change values of data in HIS handler. For other cases that is not needed and will just make overhead. 

Is it more reasonable to "say" developers who want to change data values in HIS handler to make deepcopy BEFORE make changes?...

deepcopy is more safe way, but I'm in a humble because I'm not sure that LIBRARY should care about it...

coder2020official:
Well, suppose the dev is adding some data, for post_process.
And this data is not being used in handler.
This will cause the data to be lost. However, if deep copy is used - changes made in the deep copy dictionary in _run_middlewares_and_handlers won't affect original data

telebot/async_telebot.py
```diff
@@ -558,7 +558,7 @@ async def _run_middlewares_and_handlers(self, message, handlers, middlewares, up
                             logger.error("It is not allowed to pass data and values inside data to the handler. Check your handler: {}".format(handler['function']))
                             return
                     else:
-                        data_copy = data.copy()
+                        data_copy = copy.deepcopy(data)
```
Tishka17, line 561: You should no copy data. It is required to pass original objects retrieved from middlewares to handlers

coder2020official:
@Badiboy sorry for the inconvenience, we might need to revert this PR
Could you please do this, since I currently can't

coder2020official:
Or maybe let's wait for me to get back to work, I'll be available to check disadvantages of this again on Sunday
Discard previous comment

Badiboy:
I think it may wait for Sunday.

Badiboy:
As I said from the very beginnig: you are trying to think about all developer's needs yourself. It not looks like a good practice, because in most cases you cannot even imagine all cases how the library is used by all hundreds of thousands of developers.

That's why I always propose as much flexibility for developers as it possible. In this case I still think it's better to leave decision about making deepcopy to developer of middleware...

PR #2363, opened 2024-08-11
Revert changes to middlewares

coder2020official:
Decided to revert the changes to middlewares.

To change data in handler, the user will need to have a dict parameter, which will obtain original dict, not a shallow copy.
Values are still passed to post_process regardless of handler.

Sorry for the inconvenience.


Badiboy (approved):

Badiboy:
It's OK

PR #2387, opened 2024-08-20
Fix: Correct order of parameters in forward_messages

gautamajay52:
## Description
fixed order of parameters in `forward_messages` of async telebot

## Checklist:


Badiboy:
Thank you!

PR #2389, opened 2024-08-21
Set answer_pre_checkout_query query_id argument type to str

Relecto:
Hi! 
Noticed a typing mistake, it should be fairly easy to fix :)

As per the documentation, in `answer_pre_checkout_query` method, `pre_checkout_query_id` should be str, not int - https://core.telegram.org/bots/api#answerprecheckoutquery

## Description
Correct type in answer_pre_checkout_query() method signature  

## Describe your tests
Pycharm does not show type warning anymore :)

Python version: 3.12

OS: MacOS, Debian

## Checklist:

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -5627,7 +5627,7 @@ class TeleBot:
 
 
     def answer_pre_checkout_query(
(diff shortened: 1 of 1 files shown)


Badiboy:
Thank you.

Issue #2390, opened 2024-08-30
Purpose of infinity_polling for async


coder2020official:
@Badiboy would calling polling func for infinity_polling in async make sense? I find infinity polling useless in async version.

Badiboy:
@coder2020official I'm not deep in Async version and never used it, thus have no oppinion on this topic.

Issue #2386, opened 2024-08-18
How to properly stop polling using AsyncTeleBot?


EugeneNeuron:
1. What version of pyTelegramBotAPI are you using?
4.22.0
2. What OS are you using?
Gnu/Linux
3. What version of python are you using?
3.10.7

I see how to gracefully stop polling using `TeleBot` class (#2098), but how to properly stop bot/polling using `AsyncTeleBot`?

coder2020official:
u close session first then stop polling

FaridMalekpour:
To gracefully stop polling when using `AsyncTeleBot` in `pyTelegramBotAPI`, you can use the `stop_polling()` method in combination with await to ensure that the bot's polling is properly stopped. Since `AsyncTeleBot` runs asynchronously, the `close_session()` method will close the bot's session but might not ensure proper cleanup of all resources if not handled correctly. Here's how you can gracefully stop the bot using `AsyncTeleBot`:

```python
import asyncio
from telebot.async_telebot import AsyncTeleBot

API_TOKEN = 'YOUR_BOT_API_TOKEN'
bot = AsyncTeleBot(API_TOKEN)

# Start polling
async def start_polling():
    await bot.polling(non_stop=True)

# Gracefully stop polling
async def stop_polling():
    await bot.stop_polling()  # This will stop the polling process
    await bot.close_session()  # Close the bot's aiohttp session

async def main():
    try:
        await start_polling()
    except (KeyboardInterrupt, SystemExit):
        print("Gracefully stopping the bot...")
        await stop_polling()

if __name__ == '__main__':
    asyncio.run(main())

```

EugeneNeuron:
Thank you @coder2020official and @FaridMalekpour for your answers.

But, sorry, maybe I don't understand something, but there's literally no such method as `stop_polling` for `AsyncTeleBot`.

```python
In [7]: bot = AsyncTeleBot(token)

In [9]: bot.stop_polling()
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In [9], line 1
----> 1 bot.stop_polling()

AttributeError: 'AsyncTeleBot' object has no attribute 'stop_polling'
```

coder2020official:
Hi, yes, I am currently working on this, hence this issue is still open. Will notify you when PR is ready

coder2020official:
You may try my PR version out if you want.
It should introduce(I hope) graceful & proper shutdown of polling.

[...]

PR #2392, opened 2024-08-31
Added stop_polling & changed way polling is run for (graceful) shutdown

coder2020official:


coder2020official:
#2390: removed infinity_polling fully for time being. Reason: useless; Not final decision though - If I get new ideas regarding it, I may bring things back, but @ badiboy final word is from you here.

#2386 somewhat implemented, not tested.

PR is not yet complete

PR #2394, opened 2024-09-02
Fixed typing for `request_name` parameter

pnwclw:
## Description
Small bug fix, according documentation type should be also `bool` as it is for `request_username` and `request_photo` parameters

Python version: 3.12

OS: macOS

## Checklist:


Badiboy:
Thank you.

PR #2399, opened 2024-09-30
Replaced deprecated reply_to_message_id with reply_parameters in custom_states examples

damascene:
## Description
I've replaced the deprecated `reply_to_message_id` with `reply_parameters` in the custom_states examples. I did that after noticing a deprecation message in the logs but couldn't see an example of usage of the new method.

On Telegram I saw this so I used it:

```python
bot.send_message(
    <CHAT_ID>,
    <TEXT>,
    reply_parameters=telebot.types.ReplyParameters(
        message_id=<QUOTED_MESSAGE_ID>,
        chat_id=<QUOTED_CHAT_ID>,
        quote=<QUOTED_TEXT>
    )
)
```

## Describe your tests
I've tested the example code after modification and it seems to have behaved as expected. 

Python version: 3.10

## Checklist:
- [ x] I added/edited example on new feature/change (if exists)
- [ x] My changes won't break backward compatibility
- [x ] I made changes both for sync and async

### DIFF
--- a/examples/asynchronous_telebot/custom_states.py
+++ b/examples/asynchronous_telebot/custom_states.py
@@ -2,6 +2,7 @@ from telebot import async_telebot, asyncio_filters, types
 from telebot.asyncio_storage import StateMemoryStorage
 from telebot.states import State, StatesGroup
 from telebot.states.asyncio.context import StateContext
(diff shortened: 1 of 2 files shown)


Badiboy:
Thank you!

PR #2405, opened 2024-10-17
Fix KeyboardButtonRequestChat request_title typing

maksimkulis:
## Description
Fix KeyboardButtonRequestChat request_title typing from str to bool

## Describe your tests
How did you test your change?
no tests needed

Python version: any

OS: any

## Checklist:


Badiboy:
Thank you!

PR #2409, opened 2024-10-31
Remove telegram API exception handling

mifizz:
Remove Telegram API exception handling

## Description
I think that telegram API exception handling is not needed at all, because **__threaded_polling()** handles all exceptions in *except Exception as e:* anyway.

The problem with API exception is that when it happens, *except apihelper.ApiException as e:* doesn't raise an exception and just stops bot normally (if you are using **polling()** without **non_stop**) and therefore I can't catch that exception when using **try... except...**.

I know that **polling()** is deprecated, but **infinity_polling()** uses **polling()** anyway, and I am also using regular **polling()**, because I am making my own logging and exception handling system.

## Describe your tests
Bot raises telegram API exceptions just like all other exceptions now.

Python version:
3.12.5

OS:
Windows

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -1234,31 +1234,6 @@ class TeleBot:
                 polling_thread.raise_exceptions()
                 self.worker_pool.raise_exceptions()
                 error_interval = 0.25
-            except apihelper.ApiException as e:
-                handled = self._handle_exception(e)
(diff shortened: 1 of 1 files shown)


Badiboy:
No way.

To discuss - raise a question in discussions.

PR #2416, opened 2024-11-15
Version bump and python upgrade

Badiboy:
## Description
Version bump and python upgrade


coder2020official:
Don't you think we are slightly hurrying with dropping py version support?

Issue #2425, opened 2024-11-29
mlink with escape=False


samvova:
Please answer these questions before submitting your issue. Thanks!

1. What version of pyTelegramBotAPI are you using?
4.12
2. What OS are you using?
Windows
3. What version of python are you using?
3.10

Please check code formatting.py
```python
def mlink(content: str, url: str, escape: Optional[bool]=True) -> str:
    return '[{}]({})'.format(escape_markdown(content), escape_markdown(url) if escape else content)
```
its mistake in case of escape=False - url will be changed to content

should be `return '[{}]({})'.format(escape_markdown(content), escape_markdown(url) if escape else url)`

Badiboy:
Looks reasonable. Please, push PR.

samvova:
> Looks reasonable. Please, push PR.

this is my first PR for a project of such scale as telebot. i was nervous to be honest. 
Thanks you all, Great team!

Badiboy:
> this is my first PR for a project of such scale as telebot

🤜🤛

PR #2426, opened 2024-11-29
Fix: change code for non-escape mode - return url

samvova:
## Description
The correction ensures that when escape is False, url is used instead of content in the formatted string.
[Issue#2425](https://github.com/eternnoir/pyTelegramBotAPI/issues/2425#issue-2705365214)

## Describe your tests
How did you test your change?
```python
from formatting import mlink

print(mlink("content", "http://example.com/test", escape=False))
#[content](http://example.com/test)

print(mlink("content", "http://example.com/test", escape=True))
#[content](http://example\.com/test)

```
Python version:
3.10
OS:
Windows 11

## Checklist:
- [] I added/edited example on new feature/change (if exists)

### DIFF
--- a/telebot/formatting.py
+++ b/telebot/formatting.py
@@ -240,7 +240,7 @@ def mlink(content: str, url: str, escape: Optional[bool]=True) -> str:
     :return: The formatted string.
     :rtype: :obj:`str`
     """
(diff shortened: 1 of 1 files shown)


Badiboy:
Thank you!

Issue #2430, opened 2024-12-06
AttributeError: 'InlineKeyboardButton' object has no attribute 'copy_text'


Archangelll:
Please answer these questions before submitting your issue. Thanks!

1. What version of pyTelegramBotAPI are you using? 4.24.0'

2. What OS are you using? ubuntu

3. What version of python are you using? 3.12.3

ERROR:  File "/home/ubuntu/.local/lib/python3.12/site-packages/telebot/types.py", line 3006, in to_dict
if self.copy_text is not None:
       ^^^^^^^^^^^^^^
AttributeError: 'InlineKeyboardButton' object has no attribute 'copy_text'

 looks like new version of pyTelegramBotApi does not support the InlineKeyboardButton object created before update (which has no copy_text attribute)

For compatibility between versions one should add to pyTelegramBotApi a code that assures that InlineKeyboardButton has the copy_text attribute and any other essential attributes so i do not get this type of error every time telebot updates and decides to add new attributes that didnt exist in old versions

Badiboy:
![image](https://github.com/user-attachments/assets/9df22c7e-b38f-45ed-b56d-f4ea4ed9335b)

copy_text field is created in __init__. 

How do you reach the situation where your InlineButtonKeyboard does not have this field?

Archangelll:
Not in old versions of telebot

When i created the markup, copy_text didint exist, and i just stored it in a shelve file for backup purposes so then after each restart of the bot it saves the markup object to a shelve file and then reloads it  when starting the bot again, so then i do not have to create the markup all over again

And markup in my bot is created by the user, so its easier to store the markup object itself (to reload it later) rather than the info to create it again after bot restarts

So what happend to me was that markup was created before copy_text update and when i stopped my bot, updated pyTelegramBotApi to new version and restarted my bot, the markup object stored in backup file had no copy_text button so i had to manually add it

Badiboy:
For my personal oppinion, storing serialized classes is noncence. I do not see any reason to support that noncence in class upgrades...

Archangelll:
Well ok but what im trying to say is that in order to upgrade telebot library ill have to restart my bot and as i do not want to lose my bot data im gonna have to save its state in some sort of database.

My bot creates custom messages with a lot of InlineKeyboardButton for a hundred users for example, and im always restarting it for bug correction/updates.
For me its easier to store the objects as it is (serialized) rather than  scrap it every time my bot restarts in order to create it again

[...]

PR #2431, opened 2024-12-06
Update types.py

Archangelll:
issue: https://github.com/eternnoir/pyTelegramBotAPI/issues/2430

## Description
Include changes

Python version: 3.12.3

OS: ubuntu

## Checklist:
- [ NO] I added/edited example on new feature/change (if exists)
- [ OK] My changes won't break backward compatibility
- [NO ] I made changes both for sync and async


Badiboy:
Looks like nonsence yet. 

May be reopenned after [issue](https://github.com/eternnoir/pyTelegramBotAPI/issues/2430) discussion.

Issue #2429, opened 2024-12-05
Error Sending Media Group with Video and Thumbnail: BufferedReader is Not JSON Serializable


hecklawert:
1. What version of pyTelegramBotAPI are you using?
4.24.0
2. What OS are you using?
It's the official Python 3.9 docker image. /etc/os-release says Debian 12.
3. What version of python are you using?
Python 3.9

Hi everyone,

I'm encountering an issue that's driving me crazy. I want to send a media group message from my bot to a channel. The media group contains a single video with a custom thumbnail.

Here’s the relevant part of my code:

```python
watermarked_video_path = f"/data/watermarked_{video.video.file_id}.mp4"
thumbnail_path = "/data/thumbnail.jpg"
test = telebot.types.InputMediaVideo(open(watermarked_video_path, 'rb'), thumbnail=open(thumbnail_path, 'rb'))
self.bot.send_media_group(chat_id, [test])
```
The last line of the code (`send_media_group`) fails with the following error:
```python
Traceback (most recent call last):
  File "/app/TGbot.py", line 141, in process_and_forward_media
    self.bot.send_media_group(chat_id, [test])
  File "/usr/local/lib/python3.9/site-packages/telebot/__init__.py", line 3355, in send_media_group
    result = apihelper.send_media_group(
  File "/usr/local/lib/python3.9/site-packages/telebot/apihelper.py", line 579, in send_media_group
    media_json, files = convert_input_media_array(media)
  File "/usr/local/lib/python3.9/site-packages/telebot/apihelper.py", line 2225, in convert_input_media_array
    return json.dumps(media), files
  File "/usr/local/lib/python3.9/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/usr/local/lib/python3.9/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/local/lib/python3.9/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/local/lib/python3.9/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type BufferedReader is not JSON serializable
2024-12-05 14:48:22,254 - ERROR - Error processing and forwarding the media: Object of type BufferedReader is not JSON serializable
```

Without any thumbnail code works as expected. The thumbnail is a valid JPEG file. As per the official Telegram documentation:

[...]

PR #2434, opened 2024-12-09
Fix issue with thumbnail in InputMedia types

hecklawert:
## Description

- Fixed an issue where `InputFile` objects in the `thumbnail` field caused JSON serialization errors.

- Modified the function `convert_input_media_array` to extract `InputFile` objects from `thumbnail` and replace them with `attach://` references.

## Describe your tests

- Tested with multiple configurations of `InputMedia` arrays, including videos and images with local files as `thumbnail`.

- Verified successful requests to the Telegram API to upload media groups with proper file attachments.

- Confirmed that JSON serialization no longer fails when `InputFile` is present.

Python version:
3.12

OS:
Debian 12

## Checklist:

Related to #2429

### DIFF
--- a/telebot/asyncio_helper.py
+++ b/telebot/asyncio_helper.py
@@ -2243,6 +2243,12 @@ async def convert_input_media_array(array):
             if media_dict['media'].startswith('attach://'):
                 key = media_dict['media'].replace('attach://', '')
                 files[key] = input_media.media
+            if 'thumbnail' in media_dict:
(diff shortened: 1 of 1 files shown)


Badiboy:
I have nothing against this PR, but cannot confirm if it is right or not.
@coder2020official May be you can?

coder2020official:
Hi, sorry took me a while to write this comment. This change looks reasonable, however, you haven't made changes for the async version. Could you do that as well?

PR #2436, opened 2024-12-13
Fix telebot tests

David256:
## Description
Fix the tests

Python version: 3.11.2

OS: Debian 12


Badiboy:
Where you got problems with tests? Do not see any on github.

David256:
No environment vars are set in the tests run by Github, I guess some tests are skipped

coder2020official:
Yes, they are skipped, I doubt it would hurt if we "fixed" them this way; these don't run so hence don't fail on git

David256:
Well. But to open up the possibility of improvements, the tests must be made as realistic as possible (with chat, id, tokens)

PR #2466, opened 2025-04-05
Add support for processing language parameter in message entities

pplulee:
## Description
Add support for processing code language parameter in `<pre>` tag under message entities

## Describe your tests
How did you test your change?

Python version: 3.12

OS: Windows 11

Send a message containing a code block to the bot, then see the result of the parsed HTML code.

image
image
## Checklist:

### DIFF
--- a/telebot/formatting.py
+++ b/telebot/formatting.py
@@ -425,15 +425,14 @@ def apply_html_entities(text: str, entities: Optional[List], custom_subs: Option
             url = "tg://user?id={0}".format(user.id)
         elif subst_type == "mention":
             url = "https://t.me/{0}".format(upd_text[1:])
-        elif subst_type == "pre":
-            if language is not None:
-                return '<pre><code class="language-{0}">{1}</code></pre>'.format(language, upd_text)
         upd_text = upd_text.replace("&", "&amp;").replace("<", "&lt;").replace(">", "&gt;")
         if not subst_type or not _subs.get(subst_type):
             return upd_text
         subs = _subs.get(subst_type)
         if subst_type == "custom_emoji":
             return subs.format(text=upd_text, custom_emoji_id=custom_emoji_id)
+        elif (subst_type == "pre") and language:
+            return "<pre><code class=\"language-{0}\">{1}</code></pre>".format(language, upd_text)
         return subs.format(text=upd_text, url=url)
 
     offset = 0
--- a/telebot/formatting.py
+++ b/telebot/formatting.py
@@ -418,13 +418,16 @@ def apply_html_entities(text: str, entities: Optional[List], custom_subs: Option
     utf16_text = text.encode("utf-16-le")
     html_text = ""
 
-    def func(upd_text, subst_type=None, url=None, user=None, custom_emoji_id=None):
+    def func(upd_text, subst_type=None, url=None, user=None, custom_emoji_id=None, language=None):
         upd_text = upd_text.decode("utf-16-le")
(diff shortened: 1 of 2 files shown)


Badiboy:
Looks fine. Did you perform tests of this update?

pplulee:
I was unable to run some of the tests, but that seems to be the original problem, not caused by these updates.

For example, in `test_telebot.py`, I'm confused as to why empty data was used. This caused the tests to fail.
image
image

pplulee:
I updated the test file with some optimisations. For the issues mentioned above, environment variables are used directly. All tests now pass.

Also, I added tests for the `apply_html_entities` function.

Badiboy:
> For example, in `test_telebot.py`, I'm confused as to why empty data was used.

You need to pass YOUR bot token there.

Badiboy:
> For the issues mentioned above, environment variables are used directly.

Not absolutely correct due to TOKEN var may not exist if nothing in environment, but better then ''...

pplulee:
> > For example, in `test_telebot.py`, I'm confused as to why empty data was used.
> 
> You need to pass YOUR bot token there.

I saw environment variables for token, but some tests used them and some didn't, so I was confused. I updated to use the same environment variables for all tests so we don't have to manually edit the test files.

Badiboy:
func is designed wrong.

```
    def func(upd_text, subst_type=None, url=None, user=None, custom_emoji_id=None, language=None):
        upd_text = upd_text.decode("utf-16-le")
        if subst_type == "text_mention":
            subst_type = "text_link"
            url = "tg://user?id={0}".format(user.id)
        elif subst_type == "mention":
            url = "https://t.me/{0}".format(upd_text[1:])
        upd_text = upd_text.replace("&", "&amp;").replace("<", "&lt;").replace(">", "&gt;")
        if not subst_type or not _subs.get(subst_type):
            return upd_text
        subs = _subs.get(subst_type)
        if subst_type == "custom_emoji":
            return subs.format(text=upd_text, custom_emoji_id=custom_emoji_id)
        elif (subst_type == "pre") and language:
                return upd_text='<code class="language-{0}">{1}</code>'.format(language, upd_text)
        return subs.format(text=upd_text, url=url)

```
Otherwise symbols like < and > will break the markdown.

pplulee:
> func is designed wrong.
> 
> ```
>     def func(upd_text, subst_type=None, url=None, user=None, custom_emoji_id=None, language=None):
>         upd_text = upd_text.decode("utf-16-le")
>         if subst_type == "text_mention":
>             subst_type = "text_link"
>             url = "tg://user?id={0}".format(user.id)
>         elif subst_type == "mention":
>             url = "https://t.me/{0}".format(upd_text[1:])
>         upd_text = upd_text.replace("&", "&amp;").replace("<", "&lt;").replace(">", "&gt;")
>         if not subst_type or not _subs.get(subst_type):
>             return upd_text
>         subs = _subs.get(subst_type)
>         if subst_type == "custom_emoji":
>             return subs.format(text=upd_text, custom_emoji_id=custom_emoji_id)
>         elif (subst_type == "pre") and language:
>                 return upd_text='<code class="language-{0}">{1}</code>'.format(language, upd_text)
>         return subs.format(text=upd_text, url=url)
> ```
> 
> Otherwise symbols like < and > will break the markdown.

Thanks for pointing out the problem. I've modified the code along the lines of your idea and added a test for special characters.

Badiboy:
Thank you!

PR #2471, opened 2025-04-13
Fix: Add thumbnail support for edit_message_media

skyblade-x:
**Description:**  

By Telegram specification, thumbnails can be used in the following methods:

- `InputMediaAnimation`
- `InputMediaDocument`
- `InputMediaAudio`
- `InputMediaVideo`

When using **send_media_group**, everything works correctly. For example:

```python
bot.send_media_group(
    chat_id=CHAT_ID,
    media=[InputMediaAudio(
        media=InputFile(file=audio_path), 
        thumbnail=InputFile(file=thumbnail_path)
    )],
)
```

*Result:*
image

Audio is successfully sent with a thumbnail.

---

However, currently **edit_message_media** does not support sending thumbnails.

Example with `edit_message_media`:  
```python
initial_message = bot.send_message(CHAT_ID, "Тестовое сообщение.")

  bot.edit_message_media(
     chat_id=CHAT_ID,
     message_id=initial_message.message_id,
     media=InputMediaAudio(
         media=InputFile(file=audio_path),
         thumbnail=InputFile(file=thumbnail_path)
     )
 )
```
This throws an exception:
image-1

The issue is caused because we try to JSON serialize a raw file, but according to Telegram's specification, the thumbnail should be sent using `attach://<file_attach_name>`, not serialized directly.

Relevant part of the spec:

> "The thumbnail should be in JPEG format and less than 200 kB in size. Thumbnails can't be reused and must be uploaded as a new file, so you can pass `attach://<file_attach_name>` if the thumbnail was uploaded using multipart/form-data."

---

**Why this fix is needed:**  
Without this fix, it is impossible to send or edit media messages with a thumbnail via `edit_message_media`, even though Telegram's API supports it.

**Polite closing:**  
Please review this PR. Feedback is welcome!

### DIFF
--- a/telebot/types.py
+++ b/telebot/types.py
@@ -6703,12 +6703,13 @@ class InputMedia(Dictionaryable, JsonSerializable):
         self.caption_entities: Optional[List[MessageEntity]] = caption_entities
         self.thumbnail: Optional[Union[str, InputFile]] = thumbnail
 
-        if isinstance(self.thumbnail, InputFile):
-            self._thumbnail_name = service_utils.generate_random_token()
-            self._thumbnail_dic = 'attach://{0}'.format(self._thumbnail_name)
-        else:
+        if service_utils.is_string(self.thumbnail):
             self._thumbnail_name = ''
             self._thumbnail_dic = self.thumbnail
+
+        else:
+            self._thumbnail_name = service_utils.generate_random_token()
+            self._thumbnail_dic = 'attach://{0}'.format(self._thumbnail_name)
 
         if service_utils.is_string(self.media):
             self._media_name = ''
@@ -6721,7 +6722,9 @@ class InputMedia(Dictionaryable, JsonSerializable):
         return json.dumps(self.to_dict())
 
     def to_dict(self):
-        json_dict = {'type': self.type, 'media': self._media_dic, 'thumbnail': self._thumbnail_dic}
+        json_dict = {'type': self.type, 'media': self._media_dic}
+        if self._thumbnail_dic:
+            json_dict['thumbnail'] = self._thumbnail_dic
         if self.caption:
             json_dict['caption'] = self.caption
         if self.parse_mode:
@@ -6736,8 +6739,12 @@ class InputMedia(Dictionaryable, JsonSerializable):
         """
         if service_utils.is_string(self.media):
             return self.to_json(), None
+        
+        media_dict = {self._media_name: self.media}
+        if self._thumbnail_name:
+            media_dict[self._thumbnail_name] = self.thumbnail
 
-        return self.to_json(), {self._media_name: self.media, self._thumbnail_name: self.thumbnail}
+        return self.to_json(), media_dict
 
 
 class InputMediaPhoto(InputMedia):
--- a/telebot/types.py
+++ b/telebot/types.py
@@ -6703,10 +6703,12 @@ class InputMedia(Dictionaryable, JsonSerializable):
         self.caption_entities: Optional[List[MessageEntity]] = caption_entities
         self.thumbnail: Optional[Union[str, InputFile]] = thumbnail
 
-        if service_utils.is_string(self.thumbnail):
+        if thumbnail is None:
+            self._thumbnail_name = ''
+            self._thumbnail_dic = None
+        elif service_utils.is_string(self.thumbnail):
             self._thumbnail_name = ''
             self._thumbnail_dic = self.thumbnail
-
         else:
             self._thumbnail_name = service_utils.generate_random_token()
             self._thumbnail_dic = 'attach://{0}'.format(self._thumbnail_name)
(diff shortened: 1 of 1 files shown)


coder2020official:
Hi, thanks for the PR, I will review it

Badiboy:
We'll check later after uprade to API 9.0

telebot/types.py
```diff
@@ -6713,7 +6721,7 @@ def to_json(self):
         return json.dumps(self.to_dict())
 
     def to_dict(self):
-        json_dict = {'type': self.type, 'media': self._media_dic}
+        json_dict = {'type': self.type, 'media': self._media_dic, 'thumbnail': self._thumbnail_dic}
```
Badiboy, line 6724: Optional values should be added with "if" section.

telebot/types.py
```diff
@@ -6729,7 +6737,7 @@ def convert_input_media(self):
         if service_utils.is_string(self.media):
             return self.to_json(), None
 
-        return self.to_json(), {self._media_name: self.media}
+        return self.to_json(), {self._media_name: self.media, self._thumbnail_name: self.thumbnail}
```
Badiboy, line 6740: The same here. Optional values should be added with "if" section.

telebot/types.py
```diff
@@ -6695,12 +6695,20 @@ class InputMedia(Dictionaryable, JsonSerializable):
...
-    def __init__(self, type, media, caption=None, parse_mode=None, caption_entities=None):
+    def __init__(self, type, media, caption=None, parse_mode=None, caption_entities=None, thumbnail=None):
         self.type: str = type
         self.media: str = media
         self.caption: Optional[str] = caption
         self.parse_mode: Optional[str] = parse_mode
         self.caption_entities: Optional[List[MessageEntity]] = caption_entities
+        self.thumbnail: Optional[Union[str, InputFile]] = thumbnail
+
+        if isinstance(self.thumbnail, InputFile):
```
Badiboy, line 6706: Better check in the same order and types as for media. if ... str else ..

Badiboy:
@skyblade-x Any plans to fix or close the PR?

skyblade-x:
Hi @Badiboy,

Apologies for the delay in responding — I didn’t mean to leave the PR hanging. Thank you for your patience and your feedback.

I’ve now made the requested changes, including aligning the thumbnail check logic with media and handling the optional _thumbnail_name as suggested. I really appreciate the clarity of your review — it helped me understand the style and expectations of the project better.

I'm glad to contribute and happy to make any further improvements if needed. Thanks again for the opportunity to work on this!

telebot/types.py
```diff
@@ -6695,12 +6695,21 @@ class InputMedia(Dictionaryable, JsonSerializable):
...
-    def __init__(self, type, media, caption=None, parse_mode=None, caption_entities=None):
+    def __init__(self, type, media, caption=None, parse_mode=None, caption_entities=None, thumbnail=None):
         self.type: str = type
         self.media: str = media
         self.caption: Optional[str] = caption
         self.parse_mode: Optional[str] = parse_mode
         self.caption_entities: Optional[List[MessageEntity]] = caption_entities
+        self.thumbnail: Optional[Union[str, InputFile]] = thumbnail
+
+        if service_utils.is_string(self.thumbnail):
```
Badiboy, line 6706: "thumbnail" is optional and may be None in spite of "media" :)
Should be:
```
  if thumbnail is None:
    self._thumbnail_name = ''
    self._thumbnail_dic = None
  elif service_utils.is_string(self.thumbnail):
  ...
```

telebot/types.py
```diff
@@ -6728,8 +6739,12 @@ def convert_input_media(self):
         """
         if service_utils.is_string(self.media):
             return self.to_json(), None
+        
+        media_dict = {self._media_name: self.media}
+        if self._thumbnail_name:
+            media_dict[self._thumbnail_name] = self.thumbnail
```
Badiboy, line 6745: image

media_dict[self._thumbnail_name] will conflict with media_dict[self._media_name] (overrite) if both media and thumbnail is set as "string".

Can we use some value in self._thumbnail_name instead of empty string - I'm not sure, should be tested.

Badiboy (changes requested):

skyblade-x:
@Badiboy, I have fixed your first review, and in my opinion, everything will work as expected.
изображение
 The condition will only set the media if it has a non-empty string value: **media_dict[self._thumbnail_name]**, only by executing the following code:
изображение

Badiboy:
> The condition will only set the media if it has a non-empty string value: **media_dict[self._thumbnail_name]**,

Yep, sorry, I missed this condition.

The issue is that if thumbnail is defined as string - thumbnail will never be processed. 
image
Right?

skyblade-x:
@Badiboy if thumbnail is defined as string - for example http link to thumbnail it will be processed by this code:
изображение

Badiboy:
ok )

telebot/types.py
chengying34748: HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Run
HKEY_LOCAL_MACHINE\Software\Microsoft\Windows\CurrentVersion\Run

PR #2484, opened 2025-05-16
Add py.typed marker file for PEP 561 compatibility

mohamadghaffari:
This file signals to type checkers that the library provides inline type hints and should be type-checked, improving integration with tools like Mypy.

## Description

This pull request introduces a `py.typed` marker file to the `telebot` package, by [PEP 561](https://peps.python.org/pep-0561/).

**Changes:**
*   Added an empty `telebot/py.typed` file.

**Purpose & Benefits:**
*   **PEP 561 Compliance:** Signals to type checkers (like Mypy, Pyright, Pytype) that `pyTelegramBotAPI` ships with inline type information and is intended to be type-checked.
*   **Improved Type Checking Experience:** Allows type checkers to discover and utilize the library's existing type hints correctly. This resolves issues for users who have strict type checking configurations enabled (e.g., Mypy's `disallow_any_unimported` or `check_untyped_defs`), as they will no longer incorrectly flag `telebot` imports as untyped.
*   **Enhanced Ecosystem Integration:** Promotes better integration with the broader Python static typing ecosystem.

The existing Hatch build configuration (`[tool.hatch.build.targets.wheel] include = ["telebot/*"]`) in `pyproject.toml` ensures this new `py.typed` file will be included in the distributed wheel.

## Describe your tests

**Local Mypy Verification:**
    *   Confirmed that without `telebot/py.typed` in my local `pyTelegramBotAPI` installation, Mypy version [1.15.0] (with `disallow_any_unimported = true` in `pyproject.toml`) reported errors for `telebot` imports (e.g., `error: Type of variable becomes "Any | None" due to an unfollowed import [no-any-unimported]`).
    *   Manually added the `telebot/py.typed` file to my local installation of `pyTelegramBotAPI`.
    *   Re-ran Mypy, and the previously mentioned errors related to `telebot` were resolved, indicating Mypy now correctly recognizes the library as typed.

Python version: [3.11]

OS: [macOS Sonoma]

## Checklist:
  * N/A - This change improves type checking integration and doesn't introduce new library features requiring examples.
  * N/A - This change affects packaging and type information for the entire library, not specific sync/async code paths.
 
**Additional Observation (Optional - For Maintainer Awareness):**
During testing, I also ran `mypy telebot/` directly on the library's source code after adding `py.typed`. This revealed several internal type consistency errors reported by Mypy. While this PR solely focuses on adding the `py.typed` marker to enable proper type checking by consumers (which it successfully does), these internal findings might interest the maintainers for future type hinting refinement within the library itself. This PR does not attempt to address those internal errors.


Badiboy:
> Signals to type checkers (like Mypy, Pyright, Pytype) that pyTelegramBotAPI ships with inline type information and is intended to be type-checked.

Not everything in pyTelegramBotAPI is typed. Will it hurt any cases?

Badiboy:
@coder2020official - I not actively use such type checkers, but I don't think it will make any bad.

mohamadghaffari:
> > Signals to type checkers (like Mypy, Pyright, Pytype) that pyTelegramBotAPI ships with inline type information and is intended to be type-checked.
> 
> Not everything in pyTelegramBotAPI is typed. Will it hurt any cases?

No, It just makes the typed parts of your library work better with type-checking tools, without negatively impacting the untyped parts for users.

coder2020official:
Shouldn't we change pyproject.toml for that to work?

mohamadghaffari:
> Shouldn't we change pyproject.toml for that to work?

You already included all telebot folder inside pyproject.toml. So, that is not required. 
include = ["telebot/*"]

coder2020official:
Well, I am not sure overall
I think we will need to leave this PR for now? I might need some more time to read and see how specifically it will affect the library.

mohamadghaffari:
> Well, I am not sure overall I think we will need to leave this PR for now? I might need some more time to read and see how specifically it will affect the library.

Ok. You can check here: https://peps.python.org/pep-0561/ . It's also possible to be added by modifying the setup script like:
```
setup(
    ...,
    package_data = {
        'foopkg': ['py.typed'],
    },
    ...,
    )
```
But just adding the file worked for me as well.

Badiboy:
> Well, I am not sure overall I think we will need to leave this PR for now? I might need some more time to read and see how specifically it will affect the library.

Everything you left "for read" is left forever. I propose to apply ))

PR #2531, opened 2025-12-06
Fix: correct type hint for message to allow None

KingRotiv:
## Description
The attribute `message` can be `None` in practice, but the type hint only allowed `Message | InaccessibleMessage`.

This PR updates it to:

    Optional[Union[Message, InaccessibleMessage]]

This prevents false positives in type checkers (Pyright, MyPy, Pylance) and reflects the real behavior.

## Describe your tests
Although the change doesn't require its own test, I ran the tests and detected no errors.

Python version: 3.13

OS: Windows 11

## Checklist:


Badiboy:
Resonable, thank you.

PR #2556, opened 2026-02-14
New (alternative) message entity parsers

Badiboy:
## Description
I now myself faced the issue for some hard-structured message entities parsing, so the time had come to rewrite apply_html_entities processor. I was not able to code it myself due to lack of time, so I had videcoded them. I added 3 test versions of message entity parsers (apply_html_entities processor)/

> apply_html_entities_ds
> apply_html_entities_gm
> apply_html_entities_cg

They can be chosen by setting
`formatting.ENTITY_PASER_MODE`
entity.

## Describe your tests
I made some draft tests, all of them are assumed to work. I'll take them to prod and check how things are going. If any of them will survive - I'll provide it to replace original one.

### DIFF
--- a/telebot/formatting.py
+++ b/telebot/formatting.py
@@ -11,6 +11,7 @@ from typing import Optional, List, Dict
 # "deepseek" - deepseek version
 # "gemini" - gemini version
 # "chatgpt" - chatgpt version
+# "coder" - @coder2020official version
 # other values - original version
 ENTITY_PARSER_MODE = None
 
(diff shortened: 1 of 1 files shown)


Badiboy:
@coder2020official Your oppinion?

Badiboy:
> Are you sure this is a good idea?

Current version is obviously not fully compliant. There can be various ways to solve it, I did what I could for now because I need it now ) 

You have alternative way - let's check it.

coder2020official:
If you insist on such structure - I guess it's okay

Badiboy:
We'll clear unnecessary methods after finalizing that one of them are fully compliant. And even remove the original. I need now the ability to switch methods to test them.

Badiboy:
Ok, wrong way for testing. I'll run them locally.

coder2020official:
I think we're hurrying with a solution. We shouldn't release this right away into production

PR #2557, opened 2026-02-14
Magic I wrote a year ago

coder2020official:


coder2020official:
@Badiboy this was what I wrote some time ago; Tested now (not rigorously) -- works; Could you test, too?

coder2020official:
Tests failed because of existing pre-check of this function; The new output seems to be giving the same effect

PR #2563, opened 2026-02-21
Apply new apply_html_entities procedure

Badiboy:
## Description
Apply new apply_html_entities procedure for MessageEntity parsing in text.

## Describe your tests
Tested on prod :)


Badiboy:
@coder2020official Your code works better then your tests! 😂😂😂😂😂

PR #2589, opened 2026-04-20
Add async_antiflood helper for AsyncTeleBot

Dramex:
`telebot.util.antiflood()` wraps a sync call in a retry loop that honours Telegram's `retry_after` parameter on 429 responses. `AsyncTeleBot` users have no equivalent, so every async bot re-implements the same loop around `asyncio.sleep` and `ApiTelegramException`.

This adds `async_antiflood()` next to the sync one in `telebot.util`. Same signature, same retry budget semantics, but awaits `function(*args, **kwargs)` and sleeps via `asyncio.sleep`. Drop-in for an async bot:

```python
from telebot.util import async_antiflood

for chat_id in chat_ids:
    await async_antiflood(bot.send_message, chat_id, text)
```

### Behaviour

Mirrors the sync version exactly:
- Non-429 `ApiTelegramException` propagates immediately, no retry.
- 429 responses sleep for the server-supplied `retry_after` seconds, then retry, up to `number_retries - 1` times.
- The final attempt runs without the 429 guard, so an exhausted budget surfaces the last exception to the caller.

### Tests

Four new tests in `tests/test_async_telebot.py`:

- returns on first success without sleeping
- single 429 retries exactly once after sleeping `retry_after` seconds
- non-429 error propagates without retrying
- after `number_retries` attempts the last 429 surfaces

All five tests in the file pass (the earlier strong-ref test from #2588 plus the four new ones).

### Notes

- `async_antiflood` imports `ApiTelegramException` and `asyncio` lazily at call time, same style the sync helper uses, so `telebot.util` keeps its import graph clean.
- Exported from `telebot.util.__all__`.

### DIFF
--- a/telebot/util.py
+++ b/telebot/util.py
@@ -638,6 +638,50 @@ def antiflood(function: Callable, *args, number_retries=5, **kwargs):
         return function(*args, **kwargs)
 
 
+async def async_antiflood(function: Callable, *args, number_retries=5, **kwargs):
+    """
+    Async counterpart to :func:`antiflood`. Awaits ``function(*args, **kwargs)``
+    in a retry loop that honours Telegram's ``retry_after`` parameter on 429
+    responses. Use with :obj:`telebot.async_telebot.AsyncTeleBot` calls.
+
+    Example:
+
+    .. code-block:: python3
+
+        from telebot.util import async_antiflood
+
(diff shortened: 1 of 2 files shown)


Badiboy:
LGTM. @coder2020official ?

PR #2599, opened 2026-06-11
Fix dead retry loop in asyncio_helper._process_request

hsn8086:
## Description

In `asyncio_helper._process_request` the `raise RequestTimeout(...)` statement sits inside the retry `while` loop:

https://github.com/eternnoir/pyTelegramBotAPI/blob/744e1d07ecd78c4cae288e19b02701753ed86244/telebot/asyncio_helper.py#L96-L113

As a result, the first network error (`aiohttp.ClientError`, timeout, etc.) always aborts the call immediately and `MAX_RETRIES` has no effect — the loop can never reach a second iteration. Async bots therefore lose outgoing calls (`send_message`, ...) on any transient connection hiccup.

This PR aligns the async helper with the opt-in retry convention that already exists in `telebot.apihelper` (`RETRY_ON_ERROR`, `RETRY_TIMEOUT`):

- `RETRY_ON_ERROR = False` (default): behaviour is unchanged — fail fast on the first network error.
- `RETRY_ON_ERROR = True`: network errors are retried up to `MAX_RETRIES` times with `RETRY_TIMEOUT` seconds between attempts; `RequestTimeout` is raised only after all attempts are exhausted.
- Errors reported by the Bot API itself (`ApiTelegramException`, `ApiInvalidJSONException`, `ApiHTTPException`) keep propagating immediately, as before.

Note for reviewers: when retries are enabled, a request whose response was lost mid-flight can be repeated, i.e. an API call may be executed twice (e.g. a message sent twice). This matches the semantics of the existing sync retry engine, and is one more reason the flag stays off by default.

## Describe your tests

How did you test your change?

- `cd tests && py.test`: 58 passed, 72 skipped. Master baseline on the same machine: 53 passed, 72 skipped — the 5 extra passes are the new unit tests added here, no regressions.
- New self-contained unit tests in `tests/test_asyncio_helper.py` (no TOKEN needed, network stubbed): success path, default fail-fast, recovery after transient errors, attempt count == `MAX_RETRIES`, immediate propagation of API errors.
- Live test against a local TCP server that drops the first N connections (real aiohttp I/O): with retries enabled the call recovers after 2 dropped connections (3 connection attempts observed on the server, `RETRY_TIMEOUT` sleeps in between), the exhaustion path raises `RequestTimeout` after exactly `MAX_RETRIES` attempts, and the default path still fails fast with a single attempt.
- Live `getMe` against api.telegram.org with the patched module and default flags — works unchanged.

Python version: 3.12.3

OS: Ubuntu Linux

## Checklist:
- [ ] I added/edited example on new feature/change (if exists) — there is no example covering the sync `RETRY_ON_ERROR`/`RETRY_TIMEOUT` flags either; happy to add one if desired
- [x] My changes won't break backward compatibility — `RETRY_ON_ERROR` defaults to `False`, preserving the current fail-fast behaviour exactly
- [x] I made changes both for sync and async — sync (`apihelper`) already has working opt-in retries; this brings the async helper to parity

### DIFF
--- a/telebot/asyncio_helper.py
+++ b/telebot/asyncio_helper.py
@@ -1,4 +1,4 @@
-import asyncio # for future uses
+import asyncio
 import ssl
 import threading
 import aiohttp
@@ -26,6 +26,11 @@ FILE_URL = None
 
 REQUEST_TIMEOUT = 300
 MAX_RETRIES = 3
+# Retrying on network errors is opt-in, mirroring telebot.apihelper:
+# when RETRY_ON_ERROR is True, requests failing with a network error are
+# repeated up to MAX_RETRIES times with RETRY_TIMEOUT seconds in between.
+RETRY_ON_ERROR = False
+RETRY_TIMEOUT = 2
 
 REQUEST_LIMIT = 50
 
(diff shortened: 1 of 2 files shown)


coder2020official:
We need to check the bot api PR first

Badiboy:
https://github.com/eternnoir/pyTelegramBotAPI/pull/2629

PR #2602, opened 2026-06-22
Added RichBlock (half) and RichText stuff

coder2020official:
**Rich Messages**

- [x] Added support for [Rich Messages](https://core.telegram.org/bots/features#rich-messages), allowing bots to send highly structured text and stream AI-generated replies with seamless rich formatting.
- [x] Added the classes [RichTextBold](https://core.telegram.org/bots/api#richtextbold), [RichTextItalic](https://core.telegram.org/bots/api#richtextitalic), [RichTextUnderline](https://core.telegram.org/bots/api#richtextunderline), [RichTextStrikethrough](https://core.telegram.org/bots/api#richtextstrikethrough), [RichTextSpoiler](https://core.telegram.org/bots/api#richtextspoiler), [RichTextDateTime](https://core.telegram.org/bots/api#richtextdatetime), [RichTextTextMention](https://core.telegram.org/bots/api#richtexttextmention), [RichTextSubscript](https://core.telegram.org/bots/api#richtextsubscript), [RichTextSuperscript](https://core.telegram.org/bots/api#richtextsuperscript), [RichTextMarked](https://core.telegram.org/bots/api#richtextmarked), [RichTextCode](https://core.telegram.org/bots/api#richtextcode), [RichTextCustomEmoji](https://core.telegram.org/bots/api#richtextcustomemoji), [RichTextMathematicalExpression](https://core.telegram.org/bots/api#richtextmathematicalexpression), [RichTextUrl](https://core.telegram.org/bots/api#richtexturl), [RichTextEmailAddress](https://core.telegram.org/bots/api#richtextemailaddress), [RichTextPhoneNumber](https://core.telegram.org/bots/api#richtextphonenumber), [RichTextBankCardNumber](https://core.telegram.org/bots/api#richtextbankcardnumber), [RichTextMention](https://core.telegram.org/bots/api#richtextmention), [RichTextHashtag](https://core.telegram.org/bots/api#richtexthashtag), [RichTextCashtag](https://core.telegram.org/bots/api#richtextcashtag), [RichTextBotCommand](https://core.telegram.org/bots/api#richtextbotcommand), [RichTextAnchor](https://core.telegram.org/bots/api#richtextanchor), [RichTextAnchorLink](https://core.telegram.org/bots/api#richtextanchorlink), [RichTextReference](https://core.telegram.org/bots/api#richtextreference) and [RichTextReferenceLink](https://core.telegram.org/bots/api#richtextreferencelink), which represent different types of rich formatted text.
- [x] Added the class [RichText](https://core.telegram.org/bots/api#richtext), which represents rich formatted text.
- [x] Added the class [RichBlockCaption](https://core.telegram.org/bots/api#richblockcaption), which represents the caption of a rich formatted text.
- [x] Added the class [RichBlockTableCell](https://core.telegram.org/bots/api#richblocktablecell), which represents a cell in a table.
- [x] Added the class [RichBlockListItem](https://core.telegram.org/bots/api#richblocklistitem), which represents an item in a list.
- [x] Added the classes [RichBlockParagraph](https://core.telegram.org/bots/api#richblockparagraph), [RichBlockSectionHeading](https://core.telegram.org/bots/api#richblocksectionheading), [RichBlockPreformatted](https://core.telegram.org/bots/api#richblockpreformatted), [RichBlockFooter](https://core.telegram.org/bots/api#richblockfooter), [RichBlockDivider](https://core.telegram.org/bots/api#richblockdivider), [RichBlockMathematicalExpression](https://core.telegram.org/bots/api#richblockmathematicalexpression), [RichBlockAnchor](https://core.telegram.org/bots/api#richblockanchor), [RichBlockList](https://core.telegram.org/bots/api#richblocklist), [RichBlockBlockQuotation](https://core.telegram.org/bots/api#richblockblockquotation), [RichBlockPullQuotation](https://core.telegram.org/bots/api#richblockpullquotation), [RichBlockCollage](https://core.telegram.org/bots/api#richblockcollage), [RichBlockSlideshow](https://core.telegram.org/bots/api#richblockslideshow), [RichBlockTable](https://core.telegram.org/bots/api#richblocktable), [RichBlockDetails](https://core.telegram.org/bots/api#richblockdetails), [RichBlockMap](https://core.telegram.org/bots/api#richblockmap), [RichBlockAnimation](https://core.telegram.org/bots/api#richblockanimation), [RichBlockAudio](https://core.telegram.org/bots/api#richblockaudio), [RichBlockPhoto](https://core.telegram.org/bots/api#richblockphoto), [RichBlockVideo](https://core.telegram.org/bots/api#richblockvideo), [RichBlockVoiceNote](https://core.telegram.org/bots/api#richblockvoicenote) and [RichBlockThinking](https://core.telegram.org/bots/api#richblockthinking), which represent different types of blocks in a rich formatted message.
- [x] Added the class [RichBlock](https://core.telegram.org/bots/api#richblock), which represents a block in a rich formatted message.
- [x] Added the class [RichMessage](https://core.telegram.org/bots/api#richmessage), which represents a rich formatted message.
- [x] Added the field rich_message to the class [Message](https://core.telegram.org/bots/api#message).
- [x] Added the class [InputRichMessage](https://core.telegram.org/bots/api#inputrichmessage), describing a rich message to send.
- [x] Added the class [InputRichMessageContent](https://core.telegram.org/bots/api#inputrichmessagecontent) and allowed it to be used as [InputMessageContent](https://core.telegram.org/bots/api#inputmessagecontent) in results of inline, guest, and Web App queries.
- [x] Added the method [sendRichMessage](https://core.telegram.org/bots/api#sendrichmessage), allowing bots to send rich messages.
- [x] Added the method [sendRichMessageDraft](https://core.telegram.org/bots/api#sendrichmessagedraft), allowing bots to stream partial rich messages.
- [x] Added the parameter rich_message to the method [editMessageText](https://core.telegram.org/bots/api#editmessagetext), allowing bots to edit rich messages.

**Join Request Queries**

- [x] Added the field supports_join_request_queries to the class [User](https://core.telegram.org/bots/api#user).
- [x] Added the field guard_bot to the class [ChatFullInfo](https://core.telegram.org/bots/api#chatfullinfo).
- [x] Added the field query_id to the class [ChatJoinRequest](https://core.telegram.org/bots/api#chatjoinrequest).
- [x] Added the method [answerChatJoinRequestQuery](https://core.telegram.org/bots/api#answerchatjoinrequestquery).
- [x] Added the method [sendChatJoinRequestWebApp](https://core.telegram.org/bots/api#sendchatjoinrequestwebapp).

**Polls**

- [x] Added the class [Link](https://core.telegram.org/bots/api#link) and the field link to the class [PollMedia](https://core.telegram.org/bots/api#pollmedia).
- [x] Added the class [InputMediaLink](https://core.telegram.org/bots/api#inputmedialink) and allowed it to be used as [InputPollOptionMedia](https://core.telegram.org/bots/api#inputpolloptionmedia).

### DIFF
--- a/telebot/__init__.py
+++ b/telebot/__init__.py
@@ -4453,7 +4453,105 @@ class TeleBot:
         """
         return apihelper.send_message_draft(
             self.token, chat_id, draft_id, text, parse_mode=parse_mode, entities=entities, message_thread_id=message_thread_id)
+    
+    def send_rich_message(
+            self, chat_id: Union[int, str],
+            rich_message: types.InputRichMessage,
+            business_connection_id: Optional[str]=None,
+            message_thread_id: Optional[int]=None,
+            direct_messages_topic_id: Optional[int]=None,
+            disable_notification: Optional[bool]=None,
+            protect_content: Optional[bool]=None,
+            allow_paid_broadcast: Optional[bool]=None,
+            message_effect_id: Optional[str]=None,
+            suggested_post_parameters: Optional[types.SuggestedPostParameters]=None,
+            reply_parameters: Optional[types.ReplyParameters]=None,
+            reply_markup: Optional[REPLY_MARKUP_TYPES]=None) -> types.Message:
+        """
+        Use this method to send rich messages. If the message contains a block with a media element,
+        then the bot must have the right to send the media to the chat. On success, the sent Message is returned.
 
+        Telegram documentation: https://core.telegram.org/bots/api#sendrichmessage
+
+        :param chat_id: Unique identifier for the target chat or username of the target bot, supergroup or channel in the format @username
+        :type chat_id: :obj:`int` or :obj:`str`
+
+        :param rich_message: The message to be sent
+        :type rich_message: :class:`telebot.types.InputRichMessage`
+
+        :param disable_notification: Sends the message silently. Users will receive a notification with no sound.
+        :type disable_notification: :obj:`bool`
+
+        :param protect_content: Protects the contents of the sent message from forwarding and saving
+        :type protect_content: :obj:`bool`
+
+        :param allow_paid_broadcast: Pass True to allow up to 1000 messages per second, ignoring broadcasting limits for a fee of 0.1 Telegram Stars per message. The relevant Stars will be withdrawn from the bot's balance.
+        :type allow_paid_broadcast: :obj:`bool`
+
+        :param message_effect_id: Unique identifier of the message effect to be added to the message; for private chats only
+        :type message_effect_id: :obj:`str`
+
+        :param suggested_post_parameters: A JSON-serialized object containing the parameters of the suggested post to send;
+            for direct messages chats only. If the message is sent as a reply to another suggested post, then that suggested post is automatically declined.
+        :type suggested_post_parameters: :class:`telebot.types.SuggestedPostParameters`
+
+        :param reply_parameters: Description of the message to reply to
+        :type reply_parameters: :class:`telebot.types.ReplyParameters`
+
+        :param reply_markup: Additional interface options. A JSON-serialized object for an inline keyboard, custom reply keyboard, instructions to remove a reply keyboard or to force a reply from the user.
+        :type reply_markup: :class:`telebot.types.InlineKeyboardMarkup` or :class:`telebot.types.ReplyKeyboardMarkup` or :class:`telebot.types.ReplyKeyboardRemove` or :class:`telebot.types.ForceReply`
+
+        :param business_connection_id: Identifier of a business connection
+        :type business_connection_id: :obj:`str`
+
+        :param message_thread_id: The thread identifier of a message from which the reply will be sent
+        :type message_thread_id: :obj:`int`
+
+        :param direct_messages_topic_id: Identifier of the direct messages topic to which the message will be sent; required if the message is sent to a direct messages chat
+        :type direct_messages_topic_id: :obj:`int`
+
+        :return: On success, the sent Message is returned.
+        :rtype: :class:`telebot.types.Message`
+        """
+        disable_notification = self.disable_notification if (disable_notification is None) else disable_notification
+        protect_content = self.protect_content if (protect_content is None) else protect_content
+
+        return types.Message.de_json(
+            apihelper.send_rich_message(
+                self.token, chat_id, rich_message, disable_notification=disable_notification, protect_content=protect_content,
+                allow_paid_broadcast=allow_paid_broadcast, message_effect_id=message_effect_id, suggested_post_parameters=suggested_post_parameters, reply_parameters=reply_parameters,
+                reply_markup=reply_markup, business_connection_id=business_connection_id, message_thread_id=message_thread_id, direct_messages_topic_id=direct_messages_topic_id)
+            )
+
+    def send_rich_message_draft(
+            self, chat_id: int,
+            draft_id: int,
+            rich_message: types.InputRichMessage,
+            message_thread_id: Optional[int]=None) -> bool:
+        """
+        Use this method to stream a partial rich message to a user while the message is being generated
+        Note that the streamed draft is ephemeral and acts as a temporary 30-second preview - once the output
+        is finalized, you must call sendRichMessage with the complete message to persist it in the user's chat. Returns True on success.
+
+        Telegram documentation: https://core.telegram.org/bots/api#sendrichmessagedraft
+
+        :param chat_id: Unique identifier for the target private chat
+        :type chat_id: :obj:`int`
+
+        :param draft_id: Unique identifier of the message draft; must be non-zero. Changes to drafts with the same identifier are animated.
+        :type draft_id: :obj:`int`
+
+        :param rich_message: The partial message to be streamed
+        :type rich_message: :class:`telebot.types.InputRichMessage`
+
+        :param message_thread_id: Unique identifier for the target message thread
+        :type message_thread_id: :obj:`int`
+
+        :return: Returns True on success.
+        :rtype: :obj:`bool`
+        """
+        return apihelper.send_rich_message_draft(
+            self.token, chat_id, draft_id, rich_message, message_thread_id=message_thread_id)
 
 
     def send_chat_action(
@@ -5114,6 +5212,42 @@ class TeleBot:
         """
         return apihelper.decline_chat_join_request(self.token, chat_id, user_id)
 
+    def answer_chat_join_request_query(self, chat_join_request_query_id: str, result: str) -> bool:
+        """
+        Use this method to process a received chat join request query. Returns True on success.
+
+        Telegram documentation: https://core.telegram.org/bots/api#answerchatjoinrequestquery
+
+        :param chat_join_request_query_id: Unique identifier of the join request query
+        :type chat_join_request_query_id: :obj:`str`
+
+        :param result: Result of the query. Must be either “approve” to allow the user to join the chat,
+            “decline” to disallow the user to join the chat, or “queue” to leave the decision to other administrators.
+        :type result: :obj:`str`
+
+        :return: True on success.
+        :rtype: :obj:`bool`
+        """
+        return apihelper.answer_chat_join_request_query(self.token, chat_join_request_query_id, result)
+
+    def send_chat_join_request_web_app(self, chat_join_request_query_id: str, web_app_url: str) -> bool:
+        """
+        Use this method to process a received chat join request query by showing a Mini App to the
+        user before deciding the outcome. Call answerChatJoinRequestQuery to resolve the join request query based on the user interaction with the Mini App.
+        Returns True on success.
+
+        Telegram documentation: https://core.telegram.org/bots/api#sendchatjoinrequestwebapp
+
+        :param chat_join_request_query_id: Unique identifier of the join request query
+        :type chat_join_request_query_id: :obj:`str`
+
+        :param web_app_url: The URL of the Mini App to be opened
+        :type web_app_url: :obj:`str`
+
+        :return: True on success.
+        :rtype: :obj:`bool`
+        """
+        return apihelper.send_chat_join_request_web_app(self.token, chat_join_request_query_id, web_app_url)
 
     def set_chat_photo(self, chat_id: Union[int, str], photo: Any) -> bool:
         """
@@ -5659,7 +5793,8 @@ class TeleBot:
             reply_markup: Optional[types.InlineKeyboardMarkup]=None,
             link_preview_options : Optional[types.LinkPreviewOptions]=None,
             business_connection_id: Optional[str]=None,
-            timeout: Optional[int]=None) -> Union[types.Message, bool]:
+            timeout: Optional[int]=None,
+            rich_message: Optional[types.InputRichMessage]=None) -> Union[types.Message, bool]:
         """
         Use this method to edit text and game messages.
 
@@ -5698,6 +5833,9 @@ class TeleBot:
         :param timeout: Timeout in seconds for the request.
         :type timeout: :obj:`int`
 
+        :param rich_message: New rich content of the message; required if text isn't specified
+        :type rich_message: :obj:`types.InputRichMessage`
+
         :return: On success, if edited message is sent by the bot, the edited Message is returned, otherwise True is returned.
         :rtype: :obj:`types.Message` or :obj:`bool`
         """
@@ -5725,7 +5863,7 @@ class TeleBot:
         result = apihelper.edit_message_text(
             self.token, text, chat_id=chat_id, message_id=message_id, inline_message_id=inline_message_id,
             parse_mode=parse_mode, entities=entities, reply_markup=reply_markup, link_preview_options=link_preview_options,
-            business_connection_id=business_connection_id, timeout=timeout)
+            business_connection_id=business_connection_id, timeout=timeout, rich_message=rich_message)
 
         if isinstance(result, bool):  # if edit inline message return is bool not Message.
             return result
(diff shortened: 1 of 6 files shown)


coder2020official:
Added RichText stuff and half of RichBlock classes

This could've been more simplified (e.g. by having text in the init of the base class), but I chose not to display and confuse people for some classes that don't have the text attribute, for example.

Long way to go, I hope I'm on the right way

coder2020official:
@badiboy I think I'm done, but I will double-check & test things tomorrow; you can, however, start reviewing the code

Badiboy:
> @Badiboy I think I'm done, but I will double-check & test things tomorrow; you can, however, start reviewing the code

Thank you for your work!! I do not think I can do it before weekend, but I'll do my best.

ranjitsingha:
@coder2020official Tried the `botapi101` branch with `AsyncTeleBot` and hit two issues using `send_rich_message`. Sharing in case it helps:

**1. Async `send_rich_message` raises `TypeError` on every call**

`AsyncTeleBot.send_rich_message` forwards `message_thread_id` to `asyncio_helper.send_rich_message`:

```python
await asyncio_helper.send_rich_message(
    self.token, chat_id, rich_message,
    business_connection_id=..., message_thread_id=message_thread_id,
    direct_messages_topic_id=..., ...)
```

but `asyncio_helper.send_rich_message(...)` has no `message_thread_id` parameter, so every call fails with:

```
TypeError: send_rich_message() got an unexpected keyword argument 'message_thread_id'
```

(The sync `apihelper.send_rich_message` has the same parameter mismatch.)

**2. Parsing the response crashes on plain-string rich text**

After working around (1), `Message.de_json` of the `sendRichMessage` result throws whenever a rich-text node is a plain string:

```
File ".../telebot/types.py", in RichText.de_json
ValueError: json_type should be a json dict or string.
```

Per the docs, `RichText` can be a plain `str` ("either a String for plain text, an Array of `RichText`, or any of the following types"), but `RichText.de_json` always calls `check_json`, which can't handle a bare string. So any successfully-sent rich message containing plain text fails to deserialize. `RichText.de_json` should pass `str` (and the array case) through instead of forcing `check_json`.

Minimal repro (async): `await bot.send_rich_message(chat_id, types.InputRichMessage(html="<b>hi</b>"))`.

coder2020official:
Hi!

Thanks for noting that--I have fixed your issue exactly; however, I am yet to address RichText being an array of RichText. I will have to test and see how it will look like. 

And I'm not sure my current implementation is safe. It seems to work, though.

coder2020official:
This should hypothetically work for arrays. Not really having the time to test--would be amazing if you could play around

ranjitsingha:
@coder2020official — your fix works. Confirmed on `dee734d`: `send_rich_message` no longer `TypeError`s, and the plain-string `RichText` parse is resolved.

One thing worth flagging separately, because I don't think it was obvious from how I first framed it (my original report was only about the **send** path): the same `RichText.de_json` bug also hits the **receive** path, where the blast radius is much larger.

When a `rich_message` arrives in an update — e.g. a user simply **forwards a rich message to the bot**, or forwards one of the bot's own rich messages back — it's parsed by `Update.de_json` → `Message.de_json` → `RichMessage.de_json`. On the async poller that runs *inside* the polling loop, and the exception is raised **before the offset is advanced**. So a single unparseable `rich_message` update:

- crashes on every `get_updates` cycle on the same update,
- never advances past it,
- **stalls the whole bot** — no updates of any type are processed until the poison update is manually skipped.

In effect it's a remotely-triggerable poller stall: anyone can send/forward rich content to any polling bot and freeze it — **even if the bot never sends rich messages itself**. That's actually how I hit it: an inbound forward, not a send.

```python
# no server needed to reproduce the crash Update.de_json would hit:
from telebot.types import RichText
RichText.de_json({"type": "bold", "text": "Detail"})   # pre-fix: JSONDecodeError
```

`7b99c65` fixes the specific string-leaf case, so the common trigger is gone. 👍 The reason I'm raising it anyway: you noted RichBlock support is still "half", and that the array handling in `dee734d` is untested / you're "not sure it's safe". Given the receive-side blast radius above, **any** rich node type that isn't yet handled will resurface the *same* full-poller stall — not just a failed send.

So it may be worth making the receive/parse path fail-safe while the block classes are being finished: if `RichMessage.de_json` can't parse something, drop it to `None` and keep the rest of the `Message` intact (bots that don't consume `rich_message` lose nothing), rather than letting the error propagate up and jam `get_updates`. That decouples "incomplete rich parser" from "bot stops entirely".

ranjitsingha:
@coder2020official — one more of the same family (like the earlier `send_rich_message` / `message_thread_id` mismatch you fixed in 7b99c65): **`send_rich_message_draft` raises `TypeError` on every call.**

`AsyncTeleBot.send_rich_message_draft` forwards `business_connection_id` to the helper:

```python
return await asyncio_helper.send_rich_message_draft(
    self.token, chat_id, draft_id, rich_message,
    message_thread_id=message_thread_id, business_connection_id=business_connection_id)
```

but `asyncio_helper.send_rich_message_draft` has no such parameter:

```python
async def send_rich_message_draft(token, chat_id, draft_id, rich_message, message_thread_id=None):
```

so every call fails with:

```
TypeError: send_rich_message_draft() got an unexpected keyword argument 'business_connection_id'
```

Minimal repro (async):

```python
await bot.send_rich_message_draft(chat_id, 1, types.InputRichMessage(html="<tg-thinking>hi</tg-thinking>"))
```

The sync `apihelper.send_rich_message_draft(token, chat_id, draft_id, rich_message, message_thread_id=None)` has the same missing parameter, so the sync `TeleBot` method hits it too.

**Suggested fix:** add `business_connection_id=None` to both the async and sync `send_rich_message_draft` helpers and include it in the payload when set (mirroring how `send_rich_message` handles it).

I worked around it by calling the helper directly, and the draft itself streams fine once that's bypassed.

coder2020official:
sendrichmessagedraft shouldn't even have business connection id

coder2020official:
I tried forwarding a rich message to the bot and there were no issues

ranjitsingha:
> I tried forwarding a rich message to the bot and there were no issues

It was fixed in dee734d

Badiboy:
@coder2020official Sorry for been so long. Hope to review today.

telebot/types.py
```diff
@@ -391,6 +391,10 @@ class ChatJoinRequest(JsonDeserializable):
     :param invite_link: Optional. Chat invite link that was used by the user to send the join request
     :type invite_link: :class:`telebot.types.ChatInviteLink`
 
+    :param query_id: Optional. Identifier of the join request query; for bots assigned to process join request only.
+        If present, then the bot must call sendChatJoinRequestWebApp or directly call answerChatJoinRequestQuery within 10 seconds.
+    :type query_id: :obj:`int`
```
Badiboy: String.

Image

telebot/types.py
```diff
@@ -403,13 +407,14 @@ def de_json(cls, json_string):
...
 
-    def __init__(self, chat, from_user, user_chat_id, date, bio=None, invite_link=None, **kwargs):
+    def __init__(self, chat, from_user, user_chat_id, date, bio=None, invite_link=None, query_id=None, **kwargs):
         self.chat: Chat = chat
         self.from_user: User = from_user
         self.date: str = date
         self.bio: Optional[str] = bio
         self.invite_link: Optional[ChatInviteLink] = invite_link
         self.user_chat_id: int = user_chat_id
+        self.query_id: Optional[int] = query_id
```
Badiboy: string

telebot/types.py
```diff
@@ -14456,12 +14504,40 @@ def de_json(cls, json_string):
             obj['venue'] = Venue.de_json(obj['venue'])
         if 'video' in obj:
             obj['video'] = Video.de_json(obj['video'])
+        if 'link' in obj:
+            obj['link'] = Link.de_json(obj['link'])
         return cls(**obj)
 
+
+class InputMediaLink(JsonDeserializable):
```
Badiboy: Propose to inherut it from InputMedia for consistency. Yes, I know that "media" field is required. I see two options: 
1) Make "media" optional for all InputMedia-s
2) Override "init" here to avoid "media" for this class and set it to "None" inside.

First option looks more correct.

telebot/types.py
```diff
@@ -14543,3 +14619,1777 @@ def de_json(cls, json_string):
...
+    Formatted date and time.
+
+    Telegram documentation: https://core.telegram.org/bots/api#richtextdatetime
+
+    :param type: Type of the rich text, always “date_time”
+    :type type: :obj:`str`
+
+    :param text: The text
+    :type text: :class:`RichText`
+
```
Badiboy, line 14871: Missed fields:

Image

telebot/types.py
```diff
@@ -14543,3 +14619,1777 @@ def de_json(cls, json_string):
...
+    :return: Instance of the class
+    :rtype: :class:`RichBlockTableCell`
+    """
+    def __init__(self, text: Optional[RichText] = None, is_header: Optional[bool] = None, colspan: Optional[int] = None, rowspan: Optional[int] = None,
+                    align: Optional[str] = None, valign: Optional[str] = None, **kwargs):
+        self.text: Optional[RichText] = text
+        self.is_header: Optional[bool] = is_header
+        self.colspan: Optional[int] = colspan
+        self.rowspan: Optional[int] = rowspan
+        self.align: Optional[str] = align
```
Badiboy: Not optional here also.

telebot/types.py
```diff
@@ -14543,3 +14619,1777 @@ def de_json(cls, json_string):
...
+
+    Telegram documentation: https://core.telegram.org/bots/api#richblockvoicenote
+
+    :param type: Type of the block, always “voice_note”
+    :type type: :obj:`str`
+
+    :param voice_note: The voice note
+    :type voice_note: :class:`Voice`
+
+    :param has_spoiler: Optional. True, if the media preview is covered by a spoiler animation
```
Badiboy: No has_spoiler

Image

telebot/types.py
```diff
@@ -14543,3 +14619,1777 @@ def de_json(cls, json_string):
...
+    :type caption: :class:`RichBlockCaption`
+
+    :return: Instance of the class
+    :rtype: :class:`RichBlockVoiceNote`
+
+    """
+    def __init__(self, voice_note: Voice, has_spoiler: Optional[bool] = None, caption: Optional[RichBlockCaption] = None, **kwargs):
+        super().__init__(type='voice_note', **kwargs)
+        self.voice_note: Voice = voice_note
+        self.has_spoiler: Optional[bool] = has_spoiler
```
Badiboy: Remove has_spoiler here also. And from init params.

telebot/types.py
```diff
@@ -14543,3 +14619,1777 @@ def de_json(cls, json_string):
...
+
+class RichBlockThinking(RichBlock):
+    """
+    A block with a “Thinking…” placeholder, corresponding to the custom HTML tag <tg-thinking>.
+    The block may be used only in sendRichMessageDraft, therefore it can't be received in messages
+
+    :param type: Type of the block, always “thinking”
+    :type type: :obj:`str`
+
+    :return: Instance of the class
```
Badiboy, line 16308: text missed.

Image

Badiboy (changes requested):

coder2020official:
@badiboy done

telebot/__init__.py
Badiboy, line 5869: Need to pass text as text=text, cause it's optional now.

telebot/types.py
```diff
@@ -16285,14 +16292,21 @@ class RichBlockThinking(RichBlock):
...
 
+    :param text: Text of the block. See https://t.me/addemoji/AIActions for examples of custom emoji,
+        which are recommended for usage in the block.
+    :type text: :class:`RichText`
+
     :return: Instance of the class
     :rtype: :class:`RichBlockThinking`
     """
-    def __init__(self, **kwargs):
+    def __init__(self, text: RichText = None, **kwargs):
```
Badiboy: Should be:

>  def __init__(self, text: RichText, **kwargs):

cause it's not optional.

telebot/types.py
```diff
@@ -7163,6 +7163,9 @@ def convert_input_media(self):
         """
         :meta private:
         """
+        if self.media is None:
```
Badiboy, line 7166: I propose to return 

> None, None 

here instead of Exception. And add proceeding of this result in all places where convert_input_media is called. There are only 13 usages, so it's not hard. If you agree - I can make it later in separate PR.

telebot/asyncio_helper.py
```diff
@@ -334,7 +334,7 @@ async def send_rich_message(
         reply_parameters=None, reply_markup=None, business_connection_id=None,  allow_paid_broadcast=None, direct_messages_topic_id=None,
         suggested_post_parameters=None, message_thread_id=None):
     method_url = r'sendRichMessage'
-    payload = {'chat_id': chat_id, 'rich_message': rich_message.to_json()}
+    payload = {'chat_id': str(chat_id), 'rich_message': rich_message.to_json()}
```
Badiboy, line 337: Non need for chat_id. 

Image

Nothing bad, may be left as str(chat_id), just inform )

Badiboy (changes requested):
2 small requests, 1 discussion, 1 info )

coder2020official:
I was lazy to address your convert_input_media suggestion, but to be honest, this function is never called when the class is InputMediaLink?

We need to release a new version ASAP, too, it's been long since the update

Badiboy:
Go )

Badiboy:
> I was lazy to address your convert_input_media suggestion

I checked this place - I was wrong.

>         if self.media is None:
>             return self.to_json(), None

will be fine here. So it will suite all usages.

PR #2615, opened 2026-08-15
Fix RichText

Badiboy:
## Description
1. It's now ABC
2. de_json with json-string was lost. Now it processes string under some heuristics.


Badiboy:
@coder2020official - please, take a fast look. Update is small.

coder2020official (approved):

Badiboy:
Damn, I include wrong files in last fix and did not noticed that... 🙈🙈🙈

PR #2630, opened 2026-09-18
Fix async retry 

Badiboy:
## Description
Fix async retry (https://github.com/eternnoir/pyTelegramBotAPI/pull/2599)


telebot/asyncio_helper.py
```diff
@@ -25,7 +25,9 @@
 FILE_URL = None
 
 REQUEST_TIMEOUT = 300
-MAX_RETRIES = 3
+RETRY_ON_ERROR = False
+RETRY_TIMEOUT = 2
+MAX_RETRIES = 15
```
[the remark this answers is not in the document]
Badiboy: RETRY_ON_ERROR is False, so user will update both if needed.

telebot/asyncio_helper.py
```diff
@@ -90,28 +92,28 @@ async def _process_request(token, url, method='get', params=None, files=None, **
...
 
     timeout = aiohttp.ClientTimeout(total=request_timeout)
-    got_result = False
-    current_try=0
+    max_attempts = max(1, MAX_RETRIES if RETRY_ON_ERROR else 1)
+    last_error = None
     session = await session_manager.get_session()
-    while not got_result and current_try<MAX_RETRIES-1:
-        current_try +=1
+    for current_try in range(1, max_attempts + 1):
```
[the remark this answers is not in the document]
Badiboy: Adequate files processing become really hard, especially for aiohttp. Let's left it "as is" for files...

telebot/asyncio_helper.py
```diff
@@ -25,7 +25,9 @@
 FILE_URL = None
 
 REQUEST_TIMEOUT = 300
-MAX_RETRIES = 3
+RETRY_ON_ERROR = False
+RETRY_TIMEOUT = 2
```
coder2020official, line 29: retry_delay?

coder2020official:
looks good i guess

telebot/asyncio_helper.py
coder2020official: I wonder why we need 15 max retries though

Badiboy: I made this equal to SYNC version. Both for naming and values. 
Naming better to keep, but values may be set as you wish. May change both to 3.

coder2020official:
3 should be fine yeah for max_retries

Badiboy:
Done?

coder2020official:
yes

Badiboy:
Thnx. I'll test the current version on prod for weekend. If it's ok - release.

